PR #1, opened 2021-08-25
change everything to use the require-in-the-middle API

bengl:
Title says it all.

Obviously this is semver major since literally the whole thing changed.


index.js
```diff
@@ -14,21 +19,78 @@ const proxyHandler = {
   }
 }
 
-exports._register =  function _register(name, namespace, set) {
+function _register(name, namespace, set, specifier) {
```
rochdev, line 34: Could this be moved to a different file and resolved relatively instead of exposing a private function?

index.d.ts
```diff
@@ -2,24 +2,78 @@
...
+export default Hook
+
+/**
+ * A hook function to be run against loaded modules. To be used with the
+ * lower-level APIs `addHook` and `removeHook`.
  * @param {url} string The absolute path of the module, as a `file:` URL string.
  * @param {exported} { [string]: any } An object representing the exported items of a module.
  */
-export type HookFunction = (url: string, exported: { [string]: any }) => void
+export type LLHookFunction = (url: string, exported: Namespace) => void
```
rochdev, line 59: This is a breaking change for TypeScript users. The new function should have another name instead.

test/README.md
```diff
@@ -0,0 +1,13 @@
+These tests have the following nomenclature:
+
+* Prefixed with `hook-` if they use the `Hook` class.
```
rochdev, line 3: Why not simply use folders?

rochdev (changes requested):

index.js
```diff
@@ -2,9 +2,14 @@
...
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
+const path = require('path')
+const parse = require('module-details-from-path')
+const { fileURLToPath } = require('url')
+
 const importHooks = [] // TODO should this be a Set?
 
 const setters = new WeakMap()
+const specifiers = new Map()
```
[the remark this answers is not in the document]
bengl: Not without future-proofing for loaders-in-threads. The only reason `_register` is even called from the generated code below is because loaders-in-threads may happen.

index.js
bengl: Yes, I was thinking that ought to be done.

index.d.ts
bengl: Okay.

index.d.ts
```diff
@@ -2,24 +2,78 @@
...
+   * loaded modules.
+   */
+  unhook(): void
+}
+
+export default Hook
+
+/**
+ * A hook function to be run against loaded modules. To be used with the
+ * lower-level APIs `addHook` and `removeHook`.
```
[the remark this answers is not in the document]
bengl: Legacy/deprecated seemed like a bit early, since the module itself was first released _last week_.

test/README.md
bengl: I can do folders.

index.js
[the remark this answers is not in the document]
bengl: Threads can't share object (or functions).

PR #2, opened 2021-09-15
add support for calling defineProperty on exports

rochdev:
This PR adds support for calling `defineProperty` on exports to allow using `shimmer` and similar implementations. It also fixes the tests that were checking if the module name contains `os` instead of checking it's exactly `os`.

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -11,6 +11,9 @@ const toHook = []
 const proxyHandler = {
   set(target, name, value) {
     return setters.get(target)[name](value)
+  },
+  defineProperty(target, property, descriptor) {
+    return setters.get(target)[property](descriptor.value)
   }
 }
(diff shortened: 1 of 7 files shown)


lib/register.js
```diff
@@ -11,6 +11,9 @@ const toHook = []
 const proxyHandler = {
   set(target, name, value) {
     return setters.get(target)[name](value)
+  },
+  defineProperty(target, property, descriptor) {
+    return setters.get(target)[property](descriptor.value)
```
bengl, line 16: Can we make this throw if `(!('value' in descriptor))`? The silent failure here could be disastrous.

bengl (changes requested):

bengl (approved):

PR #9, opened 2021-12-02
support extensionless entrypoints

bengl:
Roughly replicates the behaviour of stock Node.js without loaders where
extensionless entrypoints are treated as CJS.

This allows IITM to work with such tools as mocha.


rochdev (approved):

bmeck (approved):
LGTM, though I'm sure we might revisit this

PR #14, opened 2022-07-04
Add TypeScript extensions to EXTENSION_RE

### LINKED ISSUE #13: Support TypeScript extensions
## Expected Behavior

I have put together an example of chain-loading `dd-trace` and `ts-node/esm` in a native ESM project without compiling TS to JS [here](https://github.com/fardjad/dd-trace-with-esm-loader). However, to make it work, I had to patch `import-in-the-middle`.

## Actual Behavior

Without the patch, MITM does not recognize TypeScript extensions (i.e. `.ts`, `.mts`, `.cts`).

  1. Clone the example repo
  1. Install the dependencies by running `npm i --ignore-scripts`
  1. Run `npm test`

## Specifications

  - Version: 1.2.1
  - Platform: N/A
  - Subsystem: N/A

fardjad:
Closes #13


.github/workflows/ci.yml
```diff
@@ -17,3 +17,5 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: npm install
       - run: npm test
+      - run: npm run test:ts
+        if: matrix.node-version == '18.x'
```
rochdev, line 21: Since the test no longer uses multiple loaders, can more versions be tested?

fardjad: Since it's using the built-in test runner that is introduced in 18.x, it can only run on 18.x. However I think it's also possible to run the same way as other tests.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Intercept imports in Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "c8 --check-coverage --lines 90 imhotap --runner test/runtest --files test/{hook,low-level,other}/*",
+    "test": "c8 --check-coverage --lines 80 imhotap --runner test/runtest --files test/{hook,low-level,other,typescript}/*",
```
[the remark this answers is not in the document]
rochdev: Why would the coverage drop? The new code only adds a few functions that should be called by the test, so if anything it should _increase_ the coverage. In any case, if that doesn't work feel free to revert to the previous approach, all I was really asking was adding more Node versions if at all possible, but otherwise we can figure out coverage later.

.github/workflows/ci.yml
rochdev: Looks like the tests are failing on 16.x.

package.json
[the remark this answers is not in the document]
fardjad: I saw some issues about the wrong coverage being reported by c8 but not sure if this is one of them.
Anyways, I separated ts tests from js tests again but still run them the same way as js tests (without enforcing 
 a minimum coverage).

.github/workflows/ci.yml
fardjad: I think it requires versions greater than or equal to [16.12.0](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V16.md#experimental-esm-loader-hooks-api).

I'll exclude 16.10.0 from the workflow step.

fardjad: @rochdev updated the PR. Please check again

rochdev (approved):

hook.mjs
```diff
@@ -4,7 +4,7 @@
 
 const specifiers = new Map()
 
-const EXTENSION_RE = /\.(js|mjs|cjs)$/
+const EXTENSION_RE = /\.(js|mjs|cjs|ts|mts|cts)$/
```
bengl, line 7: Could you add a comment here indicating that this is temporary until we figure out a better way of supporting arbitrary extensions?

bengl (changes requested):

rochdev:
@fardjad LGTM

fardjad:
@rochdev Great! Thanks. 

So I think [this](https://github.com/DataDog/dd-trace-js/issues/2190) issue can be closed by releasing a new version of this library and updating the dependencies of `dd-trace-js`?

PR #21, opened 2022-08-30
Fix hook in node 16.17.0

uurien:
`file:node:something` was a invalid url and the last node v16.17.0 version throws an error if the URL is not valid.

### DIFF
--- a/hook.mjs
+++ b/hook.mjs
@@ -48,7 +48,9 @@ function addIitm (url) {
   if (NODE_MAJOR === 17) {
     return urlObj.protocol !== 'file:' ? 'file:' + urlObj.href :  urlObj.href
   }
-  return (urlObj.protocol !== 'file:' && urlObj.protocol !== 'node:') && NODE_MAJOR < 18 ? 'file:' + urlObj.href :  urlObj.href
+  return urlObj.protocol !== 'file:' && urlObj.protocol !== 'node:' && NODE_MAJOR < 18
+    ? 'file:' + urlObj.href
+    : urlObj.href
 }
 
 export async function resolve (specifier, context, parentResolve) {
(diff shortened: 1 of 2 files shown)


hook.mjs
```diff
@@ -45,12 +45,14 @@ function deleteIitm (url) {
 function addIitm (url) {
   const urlObj = new URL(url)
   urlObj.searchParams.set('iitm', 'true')
-  return urlObj.protocol !== 'file:' && NODE_MAJOR < 18 ? 'file:' + urlObj.href :  urlObj.href
+  if (NODE_MAJOR === 17) {
+    return urlObj.protocol !== 'file:' ? 'file:' + urlObj.href :  urlObj.href
+  }
+  return (urlObj.protocol !== 'file:' && urlObj.protocol !== 'node:') && NODE_MAJOR < 18 ? 'file:' + urlObj.href :  urlObj.href
```
rochdev, line 51: This would be easier to read on multiple lines.

rochdev:
```suggestion
  return (urlObj.protocol !== 'file:' && urlObj.protocol !== 'node:') && NODE_MAJOR < 18
    ? 'file:' + urlObj.href
    : urlObj.href
```

rochdev (changes requested):
Can this be tested? Since it causes a crash, we should definitely make sure that the issue is actually fixed and won't come back.

Also, it looks like the issue is with Node 16.17 but this focuses only on major versions. How is the original issue addressed?

uurien:
> Can this be tested? Since it causes a crash, we should definitely make sure that the issue is actually fixed and won't come back.

Yes, I've added a build for `16.17.0` to prevent it from breaking again

> Also, it looks like the issue is with Node 16.17 but this focuses only on major versions. How is the original issue addressed?

Yes, the issue is with node 16.17. But the fix doesn't break the solution with other versions. Node 16.17.0 starts to do some checks with the `url` and it doesn't support `file:node:async_hooks` like url.

rochdev (approved):

PR #22, opened 2022-09-01
Fix module interception on windows

CarlesDD:
This PR fixes the bug that occurs in Windows where non-internal module loading is not intercepted due to the platform-specific path segment separator.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
       strategy:
         matrix:
-          node-version: [ 12.x, 14.x, 16.10.0, 16.17.0, 16.x, 17.x, 18.5.0, 18.x ]
+          node-version: [ 12.x, 14.x, 16.10.0, 16.17.0, 16.x, 18.5.0, 18.x ]
 
       steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -23,3 +22,21 @@ jobs:
       - run: npm test
       - run: npm run test:ts
         if: (matrix.node-version != '12.x' && matrix.node-version != '14.x' && matrix.node-version != '16.10.0')
+
+  build-win:
+      runs-on: windows-latest
+
+      strategy:
+        matrix:
+          node-version: [ 12.x, 14.x, 16.10.0, 16.17.0, 16.x, 18.5.0, 18.x ]
```
rochdev, line 31: Why was Node 17 removed since it's supported by Windows?

CarlesDD: The following error is thrown when using the flag `--experimental-loader` on Node v17.6.0 onwards in Windows:

`TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute`

The provided value is a valid and absolute File URL path in windows.

PR #23, opened 2022-09-12
Fix node 16.15 and 16.16

uurien:


package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Intercept imports in Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "c8 --check-coverage --lines 90 imhotap --runner test/runtest --files test/{hook,low-level,other}/*",
-    "test-win": "c8 --check-coverage --lines 90 imhotap --runner test\\runtest.bat --files test/{hook,low-level,other}/*",
+    "test": "c8 --check-coverage --lines 89 imhotap --runner test/runtest --files test/{hook,low-level,other}/*",
```
uurien, line 7: I hate this PR...
I have reduced the coverage number, because I have added some NODE_VERSION related code, it means that it is code that can be played in some node version and the coverage can't increase in that jobs.

Qard (approved):

rochdev (approved):

package.json
rochdev: It's possible to merge reports so that doesn't happen, but that's out of the scope of this PR.

Qard (approved):

PR #28, opened 2023-06-02
add Node.js v20 support

bengl:
Since Node.js v20 moves loaders to a separate thread, we can no longer depend on loading the modules in the loader to get the exports. This is needed to add our mutable proxy. For Node.js 20, exports are now retrieved via parsing.

To reduce startup overhead on older versions of Node.js, the previous method of getting exports is used, avoiding loading and using parsers.


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.10.0, 16.16.0, 16.17.0, 16.x, 17.x, 18.5.0, 18.x]
+        node-version: [12.x, 14.x, 16.10.0, 16.16.0, 16.17.0, 16.x, 17.x, 18.5.0, 18.x, 20.x]
```
Qard, line 13: Isn't 17.x long dead now? Do we still need to support that?

lib/get-esm-exports.js
```diff
@@ -0,0 +1,93 @@
...
+        break
+      case 'ExportAllDeclaration':
+        if (node.exported) {
+          exportedNames.add(node.exported.name)
+        } else {
+          exportedNames.add('*')
+        }
+        break
+      default:
+        throw new Error('unrecognized export type: ' + node.type)
```
tlhunter, line 37: I do wonder, if node adds a new export type at some point, is it safer to warn and ignore it?

tlhunter: Actually, this is probably tied to an enumeration of values built into this specific version of Acorn and therefor isn't likely to change.

lib/get-exports.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+const getEsmExports = require('./get-esm-exports.js')
+const { parse: getCjsExports }= require('cjs-module-lexer')
```
tlhunter, line 4: I'm surprised `}=` slipped through the linter

tlhunter: Oh, looks like `lib/` is a new directory and needs to be added to the linter rules?

hook.js
```diff
@@ -14,6 +14,14 @@ const NODE_MINOR = Number(NODE_VERSION[1])
 
 let entrypoint
 
+if (NODE_MAJOR >= 20) {
+  getExports = require('./lib/get-exports.js')
+} else {
+  getExports = async function (url) {
+    return Object.keys(await import(url))
+  }
```
Qard, line 22: Could also be written like this to slightly reduce promise activity and be a bit more minimal, but it's probably fine as-is too.

```suggestion
  getExports = (url) => import(url).then(Object.keys)
```

package.json
```diff
@@ -37,6 +36,9 @@
     "typescript": "^4.7.4"
   },
   "dependencies": {
+    "acorn": "^8.8.2",
```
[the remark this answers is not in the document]
tlhunter: Compressed, the files take up 133KB, which would be the approx change to a Lambda bundle.

test/fixtures/esm-exports.txt
tlhunter, line 1: Is this a text file to avoid the linter?

lib/get-exports.js
```diff
@@ -0,0 +1,51 @@
...
+  // `parentLoad` gives us the possibility of getting the source
+  // from an upstream loader. This doesn't always work though,
+  // so later on we fall back to reading it from disk.
+  const parentCtx = await parentLoad(url, context)
+  let source = parentCtx.source
+  const format = parentCtx.format
+
+  // TODO support non-node/file urls somehow?
+  if (format === 'builtin') {
+    // Builtins don't give uys the source property, so we're stuck
```
Qard, line 22:
```suggestion
    // Builtins don't give us the source property, so we're stuck
```

test/get-esm-exports/v20-get-esm-exports.js
tlhunter, line 1: Clever stuff

tlhunter (commented):
Overall this looks pretty good but we should have @qard or @rochdev weigh in first.

lib/get-exports.js
```diff
@@ -0,0 +1,51 @@
...
+
+  if (format === 'module') {
+    return getEsmExports(source)
+  }
+  if (format === 'commonjs') {
+    return addDefault(getCjsExports(source).exports)
+  }
+
+  // At this point our `format` is either undefined or not known by us. Fall
+  // back to parsing as ESM/CJS.
```
Qard, line 40: Could it not be wasm? What happens in that case?

lib/get-esm-exports.js
Qard, line 37: Would throwing interrupt the loading entirely? Perhaps we should do console.error instead?

test/fixtures/esm-exports.txt
```diff
@@ -0,0 +1,32 @@
...
+export default expression; //| default
+export default function functionName() { /* … */ } //| default
+export default class ClassName { /* … */ } //| default
+export default function* generatorFunctionName() { /* … */ } //| default
+export default function () { /* … */ } //| default
+export default class { /* … */ } //| default
+export default function* () { /* … */ } //| default
+
+// Aggregating modules
+export * from "module-name"; //| *
```
Qard, line 26: A `*` expands in-context to all the names of the thing being re-exported, but we're just calling it `*` here. Do we need to identify the _actual_ names somehow or would we just consider it the responsibility of the user to patch it at the source rather than at the point of re-export? If so, what even is the function of capturing a `*` export here? Can we just mark it as no named export to see? Or do you think we _need_ to mark the `*` for some reason?

Qard (approved):
Some questions for you, but otherwise LGTM.

lib/get-exports.js
tlhunter: Whoops, wrong repo.

lib/get-exports.js
bengl: AFAIK that's not in Node.js 20, so we'll cross that bridge when we come to it.

lib/get-esm-exports.js
bengl: Yeah as @tlhunter suggested, maybe that or `process.emitWarning()`

bengl: It'll change if:

1. There's an AST type that can go here that we're not yet testing for.
2. An AST type is added as the language evolves.

Ignoring is probably the safest in terms of not breaking user code. A warning might help indicate what the issue is.

.github/workflows/ci.yml
bengl: Yes. For now.

test/fixtures/esm-exports.txt
bengl: Yeah because it's a re-export in totality, it's not particularly useful. I just included it here for completion's sake. I think it's important to support it here in case we want or need to support it later.

tlhunter (approved):

test/fixtures/esm-exports.txt
Qard: Fair. More just a question than anything really. I don't think there's any _harm_ to it being there, just doesn't do the expanding so as long as we're fine with that then seems acceptable to me. 👍🏻

bengl: Yeah we're fine with that for IITM usage, since it makes more sense to patch the original than the re-exported one.

lib/get-exports.js
Qard: It's behind a flag, but it _is_ there. There's also json modules.

If I understand right, if we don't return _anything_ from the loader then it will skip it and use the default/built-in logic, correct? So just not handling formats we don't specifically support here is fine?

PR #35, opened 2023-11-06
Starting with to rewrite in ESM

uurien:


hook.js
```diff
@@ -2,6 +2,9 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
+const dc = require('diagnostics_channel')
```
tlhunter, line 5: Can we use `dc-polyfill` here? Should help in case anyone uses IITM with older versions of Node.js.

uurien: I guess that we can, yes. I'll change it, but for now this is just a PoC :-D

PR #36, opened 2023-11-07
Fix loader cyclical dependency bug

khanayan123:
fixes loader cyclical dependency bug by forgoing module wrapping of file and embedding file contents within the patch instead. 

To ensure the patch still works when embedding file content an AST is generated to make modifications to the file contents. The general flow is file content -> AST -> update const named exports to let & add a name to any non named default export -> regenerate file contents using updated AST -> embed the file contents in the patch.

Recast (https://github.com/benjamn/recast) is used as the AST transformer & code generator of choice since it seems to be great at maintaining original file content structure & thus subsequently maintaining original file stack trace. Although the use of source maps might be better.

This PR aims to solve this issue: https://github.com/DataDog/dd-trace-js/issues/3595


khanayan123:
Going to work on getting the CI to pass, but in the meanwhile wanted to open the PR to get any opinions on the general approach

khanayan123:
If needed we can also use https://swc.rs/ instead which is written in RUST and aims to be faster then recast/acorn.

We can draw inspiration from: https://github.com/DataDog/dd-native-iast-rewriter-js/

test/fixtures/useEnv.mjs
```diff
@@ -0,0 +1,5 @@
+import {env} from "./env.mjs";
```
bengl, line 1: Rename this file to use-env.mjs

test/hook/v16-declaration-exports.mjs
```diff
@@ -0,0 +1,54 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2.0 License.
```
bengl, line 1: This, and all other test files here, only test from v16 and up. What's the reason for this?

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Intercept imports in Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "c8 --check-coverage --lines 85 imhotap --runner 'node test/runtest' --files test/{hook,low-level,other,get-esm-exports}/*",
+    "test": "c8 --check-coverage --lines 80 imhotap --runner 'node test/runtest' --files test/{hook,low-level,other,get-esm-exports}/*",
```
bengl, line 7: Why did the coverage drop?

hook.js
```diff
@@ -138,6 +154,41 @@ set.${n} = (v) => {
...
+        fileContents = outPut.code
+        exportAlias = outPut.exportAlias
+      } catch (parseError) {
+        console.error(`Tried AST parsing ${realPath}, got error: ${parseError}`)
+        return parentGetSource(url, context, parentGetSource)
+      }
+      const src = `${fileContents}
+import { register as DATADOG_REGISTER_FUNC } from '${iitmURL}'
+{
+const set = {}
```
bengl, line 178: Indent from here to line 187.

hook.js
```diff
@@ -20,6 +23,12 @@ if (NODE_MAJOR >= 20) {
   getExports = (url) => import(url).then(Object.keys)
 }
 
+if (NODE_MAJOR >= 16) {
```
bengl, line 26: Both `getEsmExports` and `getExports` need to have a variable declared with `let`, like `entrypoint` is on line 18.

hook.js
```diff
@@ -107,12 +116,16 @@ function createHook (meta) {
     ) {
       return url
     }
-
+    
+    // on Node 16.12.0, url.format is undefined for .mjs files, so explicitly set it to 'module'
+    if (url.format === undefined && EXTENSION_MJS_RE.test(url.url)) {
```
bengl, line 121: You might also need to check if `"type": "module"` in package.json. Actually, you might not. Can you clarify the comment? What are the conditions under which it's undefined, exactly? The documentation doesn't seem to indicate this.

hook.js
```diff
@@ -138,6 +154,41 @@ set.${n} = (v) => {
 `).join('\n')}
 register(${JSON.stringify(realUrl)}, namespace, set, ${JSON.stringify(specifiers.get(realUrl))})
 `
+      } 
+    }
+    else if (NODE_MAJOR >= 16 && context.format === 'module') {
```
bengl, line 159: This should be `} else if (...` on line 158.

bengl (changes requested):

hook.js
khanayan123: on Node's 16.0.0-16.12.0, url.format is undefined for the cyclical dependency test files ./test/fixtures/a.mjs & ./test/fixtures/b.mjs 
so this explicitly sets format to 'module' for files with a .mjs extension so that they can go through the ast parsing patch for Node >= 16

bengl: Okay, what happens when `"type": "module"` and the file extension is just `.js`, but otherwise it's the same test/example?

hook.js
```diff
@@ -138,6 +154,41 @@ set.${n} = (v) => {
...
+      let fileContents
+      try {
+        fileContents = fs.readFileSync(fileURLToPath(url), 'utf8')
+      } catch (parseError) {
+        console.error(`Had trouble reading file: ${fileContents}, got error: ${parseError}`)
```
bengl, line 164: Don't use `console.error` in non-test code. If logging is handy here, use [`debuglog`](https://nodejs.org/dist/latest-v21.x/docs/api/util.html#utildebuglogsection-callback)

lib/get-esm-exports.js
```diff
@@ -14,81 +17,187 @@ function warn (txt) {
...
+      }
+  
+      this.traverse(path)
+    },
+  
+    visitExportDefaultDeclaration(path) {
+      const node = path.node
+      const iitmRenamedExport = 'iitmRenamedExport'
+  
+      if ((node.declaration.type === 'ObjectExpression' || node.declaration.type === 'ArrayExpression' || node.declaration.type === 'Literal') && generate) {
```
bengl, line 53: Avoid doing the object access 3 times like this. Consider stashing `node.declaration.type` in a variable, or doing something like: `['ObjectExpression', 'ArrayExpression', 'Literal'].includes(node.declaration.type)`

lib/get-esm-exports.js
```diff
@@ -14,81 +17,187 @@ function warn (txt) {
...
+      for (const varDecl of declaration.declarations) {
+        if (declaration.kind === 'const') {
+          declaration.kind = 'let'
+        }
+        parseVariableDeclaration(varDecl, exportAlias)
       }
       break
     case 'ClassDeclaration':
-      exportedNames.add(node.declaration.id.name)
+      exportAlias[declaration.id.name] = declaration.id.name
```
bengl, line 154: This is now the same case as `FunctionDeclaration`, so they can be handled together.

hook.js
bengl: You can also use `warn`/`process.emitWarning` as is done elsewhere.

test/hook/v16-declaration-exports.mjs
```diff
@@ -0,0 +1,54 @@
...
+   name3 as n3, 
+   name4 as n4, 
+   functionName as fn, 
+   ClassName as cn, 
+   generatorFunctionName as gfn, 
+   name5 as n5, 
+   bar as n6, 
+   name7 as n7, 
+   name8 as n8
+  } from '../fixtures/export-types/declarations.mjs'
```
Qard, line 17:
```suggestion
import {
  name1 as n1,
  name2 as n2, 
  name3 as n3, 
  name4 as n4, 
  functionName as fn, 
  ClassName as cn, 
  generatorFunctionName as gfn, 
  name5 as n5, 
  bar as n6, 
  name7 as n7, 
  name8 as n8
} from '../fixtures/export-types/declarations.mjs'
```

test/hook/v16-list-exports.mjs
```diff
@@ -0,0 +1,33 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+
+import Hook from '../../index.js'
+import n, { name1 as n1,
+   name2 as n2, 
+   name3 as n3, 
+   name4 as n4, 
+   name5 as n5, 
+   "name" as n6,
+  } from '../fixtures/export-types/list.mjs'
```
Qard, line 12:
```suggestion
import n, {
  name1 as n1,
  name2 as n2, 
  name3 as n3, 
  name4 as n4, 
  name5 as n5, 
  name as n6,
} from '../fixtures/export-types/list.mjs'
```

test/other/v16-cyclical-dependency.mjs
```diff
@@ -0,0 +1,24 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2.0 License.
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+
+import { spawn } from 'child_process';
+import { strictEqual } from 'assert';
+
+const nodeProcess = spawn('node', ['--loader', './hook.mjs', './test/fixtures/a.mjs']);
```
Qard, line 8:
```suggestion
const nodeProcess = spawn('node', [
  '--loader',
  './hook.mjs',
  './test/fixtures/a.mjs'
]);
```

test/other/v16-reassign-let.mjs
```diff
@@ -0,0 +1,24 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2.0 License.
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+
+import { spawn } from 'child_process'
+import { strictEqual } from 'assert'
+
+const nodeProcess = spawn('node', ['--loader', './hook.mjs', './test/fixtures/reassign-let.mjs'])
```
Qard, line 8:
```suggestion
const nodeProcess = spawn('node', [
  '--loader',
  './hook.mjs',
  './test/fixtures/reassign-let.mjs'
])
```

package.json
khanayan123: Coverage is going to be low on versions that aren't running some of the new code and thus tests will always fail

test/hook/v16-declaration-exports.mjs
khanayan123: Alot of the new tests contain syntax that only works in node >= 16

bengl (approved):
@khanayan123 can you clarify why this is a semver-major?

lib/get-esm-exports.js
```diff
@@ -1,95 +1,200 @@
...
+          declaration: {
+            type: 'Identifier',
+            name: iitmRenamedExport,
+          },
         }
-        break
-      default:
-        warn('unrecognized export type: ' + node.type)
+        path.insertAfter(newExportDefaultDeclaration)
+      } else {
```
Qard, line 76: Are you certain this branch is and will continue to be valid for _all_ exceptions to the `if` condition above? Would this be safer as an `else if (condition)`?

lib/get-esm-exports.js
```diff
@@ -1,95 +1,200 @@
...
-  return Array.from(exportedNames)
+
+  return Object.keys(exportAlias)
 }
 
-function parseDeclaration (node, exportedNames) {
-  switch (node.declaration.type) {
+function convertExportSpecifierToLet(exportSpecifierNames, ast) {
+  recast.visit(ast, {
+    visitVariableDeclaration(path) {
```
Qard, line 123: Visit is recursive, is it not? So wouldn't this mutate any const declarations in nested functions too? Might be better, especially for perf, to only modify top-level declarations.

Qard (approved):

lib/get-esm-exports.js
khanayan123: I would assume so because I have added a test for all the possible default exports based off of this:

https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export

khanayan123: but it might be safer to do an else if

Qard (approved):

PR #41, opened 2023-12-01
Extensionless support in 18.19

bengl:
Node.js 18.19 backported the extensionless supported added ion Node.js 21.0.0 and 20.10.0, so the test needed to be fixed for that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,25 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -10,8 +10,25 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.10.0, 16.16.0, 16.17.0, 16.x, 17.x, 18.5.0, 18.x, 20.9.0, 20.x, 21.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version:
```
tlhunter, line 13: Starting to look like dc-polyfill!

What about going further with this:

```yaml
- 12.0
- 12.x
- 14.0
- 14.x
```

PR #42, opened 2023-12-04
Proxy default properties as well as named

jsumners-nr:
This PR is a follow up to a discussion in Slack. As it is, IITM hooks will easily hook a "named" export from a CJS module but extra effort is required to hook the same exports that are defined on the "default" export. Prior to this PR, the lines 20 and 23 in the included test case will fail.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -128,6 +128,9 @@ ${exportNames.map((n) => `
 let $${n} = namespace.${n}
 export { $${n} as ${n} }
 set.${n} = (v) => {
+  if ('${n}' !== 'default' && namespace.default['${n}']) {
+    namespace.default['${n}'] = v
+  }
   $${n} = v
   return true
 }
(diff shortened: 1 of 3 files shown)


jsumners-nr:
I'm not sure who to request a review from, but @bengl I am familiar with you, so: tag, you're it.

hook.js
```diff
@@ -128,6 +128,9 @@ ${exportNames.map((n) => `
 let $${n} = namespace.${n}
 export { $${n} as ${n} }
 set.${n} = (v) => {
+  if ('${n}' !== 'default' && namespace.default['${n}']) {
+    namespace.default['${n}'] = v
```
bengl, line 132: I think you need to make sure that this only applies to CJS modules. This could easily be incorrect for ESM modules.

bengl (changes requested):

hook.js
jsumners-nr: Noted. I'm putting this PR on hold until we have resolution on my others. Even if this one were merged, I'd still have to proxy the returned proxy because I need a symbol to detect the proxy. So I want to write up a case for that and discuss it.

PR #43, opened 2023-12-07
Support `export * from 'module'` ESM syntax

jsumners-nr:
This change adds support for modules that export entities through the `export * from 'module'` ESM syntax. This resolves issue #31.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -2,7 +2,7 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
-const { randomBytes } = require('node:crypto')
+const { randomBytes } = require('crypto')
 const specifiers = new Map()
 const isWin = process.platform === "win32"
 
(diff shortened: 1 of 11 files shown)


jsumners-nr:
Putting this in draft because there is another case that needs to be handled.

hook.js
```diff
@@ -121,23 +195,53 @@ function createHook (meta) {
 
   const iitmURL = new URL('lib/register.js', meta.url).toString()
   async function getSource (url, context, parentGetSource) {
+    const imports = []
+    const namespaceIds = []
+
     if (hasIitm(url)) {
       const realUrl = deleteIitm(url)
       const exportNames = await getExports(realUrl, context, parentGetSource)
+      const setters = []
```
bengl, line 204: This whole block seems to be repeated in `processStarExport()`. Can this be DRYed? There's enough happening here that could get modified in the future, so I'd have this logic only written once.

bengl (changes requested):

hook.js
jsumners-nr: It was a bit tricky, but I managed it. Let me know what you think.

bengl (approved):

PR #53, opened 2023-12-21
Fix duplicate default exports

jsumners-nr:
Ref: https://github.com/newrelic/node-newrelic/issues/1920

<strike>👋 my last "fix" (#43) broke in a new way thanks to the helpful ESM allowance of exporting `default` multiple times. This PR resolves the issue. I suspect there could be further problems with the exports since I have no idea how exporting `default` multiple times is meant to be interpreted, but I suspect any such issues will be an extreme edge case.

The only other thing I can think of to do here is to somehow track that we already have a `default` export and munge the names of any subsequent ones.</strike>

-----

I have since learned that ESM doesn't actually allow exporting `default` multiple times. Transitive `default` exports get mapped to some other name for the module that has imported them, and only the directly imported `default` gets exported. Still, we have to handle figuring out which `default` is the right one and construct our shim namespace accordingly.

-----

2023-01-03: this can only be supported on Node versions where we parse the modules with an AST parser and interpret the code manually. So I have updated the test suite for this feature to only run on such versions.


jsumners-nr:
Moving back to draft as the issue isn't quite solved. It is possible the wrong default gets exported. In the case of `got`, it is possible the "request" default gets exported instead of the main `got` API.

hook.js
```diff
@@ -194,21 +267,60 @@ function createHook (meta) {
...
+      // and remap them to a name based on the module name. This prevents us
+      // from overriding the top-level module's (the one actually being imported
+      // by some source code) default export when we merge the namespaces.
+      const renamedDefaults = setters
+        .map(s => {
+          const matches = /let \$(.+) = (\$.+)\.default/.exec(s)
+          if (matches === null) return
+          return `_['${matches[1]}'] = ${matches[2]}.default`
+        })
+        .filter(s => s)
```
bengl, line 290: We'll probably want to eventually rework this for performance, but for now, can you can you add a double-exclamation to make it clear that it's filtering on truthiness, and therefore eliminating `undefined` holes?
```suggestion
        .filter(s => !!s)
```

bengl (approved):

hook.js
jsumners-nr: In my opinion, the `!!` construct is very difficult to read. Would a comment suffice?

PR #54, opened 2024-01-03
Add CI action to upload html coverage report

jsumners-nr:
So we can easily see what the results of code coverage check is.


jsumners-nr:
You can see what a report looks like from this PR:

image

Issue #57, opened 2024-01-23
`@@toStringTag` property not present on modules passed to `hookFn` 


pichlermarc:
## Expected Behavior

Using `Hook`, the module passed to `hookFn` includes `@@toStringTag` property as was the case up until `import-in-the-middle@1.7.1`

## Actual Behavior

Using `Hook`, the module passed to `hookFn` does not include the `@@toStringTag` property.

## Steps to Reproduce the Problem

  1. `npm init`
  2. `npm install --save-exact koa@2.15.0` (used as an example)
  3. `npm install --save-exact import-in-the-middle@1.7.3`
  4. create `index.mjs`
  ```mjs
// index.mjs
import Hook from 'import-in-the-middle'
import * as koa from 'koa';

Hook(['koa'], (exported, name, baseDir) => {
    // Expect "Module"
    if(exported[Symbol.toStringTag] !== "Module"){
        throw new Error('Expected module')
    }
})

console.log('Everything as expected');
  ```
 5. `node --loader=import-in-the-middle/hook.mjs index.mjs` **This fails**

However, doing the same with `import-in-the-middle@1.7.1` works as expected:

 7. `npm install --save-exact import-in-the-middle@1.7.1`
 8. `node --loader=import-in-the-middle/hook.mjs index.mjs` **This succeeds**

## Additional info

I'm not certain that this is safe, but adding this to the generated code in `hook.js` after [L312](https://github.com/DataDog/import-in-the-middle/blob/c3c2c52c1915b47994af59d507c59029c1f1fae9/hook.js#L312) includes the property from the primary namespace and seems to fix the issue.   

```
for (const k of Object.getOwnPropertySymbols(primary)) {
   _[k] = primary[k]
 }
```

## Specifications

  - Version: Node.js v18.19.0
  - Platform: Ubuntu 22.04

PR #64, opened 2024-03-18
added module property symbols

ida613:


khanayan123 (commented):
Could we add a test for this?

JamieDanielson:
To link the two for easy finding, seems this is intended to close #57 as it matches the suggested change

trentm:
I opened an alternative PR at #66

trentm:
#66 has gone in, so this could probably be closed now.

PR #65, opened 2024-03-25
WIP: add optional argument to specify which modules to patch

ida613:
Add the option to specify a list of modules to patch through env var MODULES_TO_PATCH. This allows IITM to ignore some files, potentially resulting in a performance gain. This also might help eliminating IITM errors since less files are being patched.
If MODULES_TO_PATCH is not specified, everything behaves as usual.

### DIFF
--- a/filtered-hook.mjs
+++ b/filtered-hook.mjs
@@ -1,0 +1,6 @@
+import { createHook } from './hook.js'
+import path from 'path'
+
+const { load, resolve, getFormat, getSource } = createHook(import.meta, new Set([path.join(import.meta.url, '../test/fixtures/export-types/default-expression-array.mjs').replace(':/', ':///')]))
+
+export { load, resolve, getFormat, getSource }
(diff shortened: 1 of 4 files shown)


khanayan123 (approved):

khanayan123:
could we add some unit tests to verify the functionality?

Qard (commented):
Why an env var vs passing the config in as a parameter to `createHook`? An env var seems a bit clumsy. 🤔

ida613:
> Why an env var vs passing the config in as a parameter to `createHook`? An env var seems a bit clumsy. 🤔

If we pass in module list as a parameter to `createHook`, the user would have to customize the hook by copying the content of hook.mjs to their own project. Putting the hook in a different directory would mess up [`isIitm`](https://github.com/DataDog/import-in-the-middle/blob/main/hook.js#L34) and [`iitmURL`](https://github.com/DataDog/import-in-the-middle/blob/main/hook.js#L265)

timfish:

We've shipped a similar feature where the excluded modules can be sent to the hook when you register it.

PR #70, opened 2024-04-17
Use updated/renamed version of acorn-import-assertions (security fix)

jackwhelpton:
This dependency appears to have been renamed since 1.9.0: observe that github.com/xtuc/acorn-import-assertions redirects to https://github.com/xtuc/acorn-import-attributes.

In 1.9.2 a security vulnerability was addressed (fully qualifying a package reference to prevent a confusion attack), which is being introduced into our codebase via this repo (by way of `dd-trace`).

Validated that all current tests still pass with this update.

### DIFF
--- a/lib/get-esm-exports.js
+++ b/lib/get-esm-exports.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { Parser } = require('acorn')
-const { importAssertions } = require('acorn-import-assertions')
+const { importAssertions } = require('acorn-import-attributes')
 
 const acornOpts = {
   ecmaVersion: 'latest',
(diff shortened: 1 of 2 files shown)


cedricvanrompay-datadog:
I had a look at https://www.npmjs.com/package/acorn-import-assertions and https://www.npmjs.com/package/acorn-import-attributes: as far as I can tell, this seems to be a legitimate change of package name, still controlled by the same maintainer.

Qard:
Yeah, the rename of the module is because the proposal for the spec was renamed.

jackwhelpton:
Any chance of a rough ETA for getting this reviewed/in, a new version cut and `dd-trace-js` updated? Our security folks are pushing us on this one and I'd like to be able to give them a rough guide on how long it'll take us to mitigate.

bengl (approved):

PR #74, opened 2024-05-06
test: add test for import attributes

chentsulin:
Add a test for import attributes on Node 20.10+.


Qard (approved):

bengl (approved):

bengl:
@chentsulin The acorn parser seems to be erroring out on tests on the 20.x builds. Can you take a look at that please?

bengl (changes requested):
Test failures need to be fixed.

timfish:

I think we support import attributes now!

Issue #63, opened 2024-01-28
Fails to resolve `exports` defined submodule when reexported by another package


isaacs:
`node_modules/has-submodule/package.json`
```json
{
  "name": "has-submodule",
  "type": "module",
  "exports": {
    "./sub": "./sub.js"
  }
}
```

`node_modules/has-submodule/sub.js`
```js
export const foo = 'bar'
console.log('loaded submodule')
cat: node_modules/reexport-submodule/sub.js: No such file or directory
```

`node_modules/reexport-submodule/package.json`
```json
{
  "name": "reexport-submodule",
  "exports": {
    ".": "./index.js"
  },
  "type": "module"
}
```

`node_modules/reexport-submodule/index.js`
```js
export * from 'has-submodule/sub'
```

`load-sub.mjs`

```js
import { foo } from 'reexport-submodule'
console.log(foo)
```

`esm-loader.mjs`

```js
export * from 'import-in-the-middle/hook.mjs'
```

## Expected Behavior

```
$ node load-sub.mjs
loaded submodule
bar
```

```
$ node --loader=./esm-loader.mjs load-sub.mjs
loaded submodule
bar
```

## Actual Behavior

```
$ node --loader=./esm-loader.mjs load-sub.mjs
(node:36249) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("./esm-loader.mjs", pathToFileURL("./"));'
(Use `node --trace-warnings ...` to show where the warning was created)

node:internal/process/esm_loader:40
      internalBinding('errors').triggerUncaughtException(
                                ^
[Error: ENOENT: no such file or directory, open '/Users/isaacs/dev/tapjs/esm-tap-repro/node_modules/reexport-submodule/has-submodule/sub'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/isaacs/dev/tapjs/esm-tap-repro/node_modules/reexport-submodule/has-submodule/sub'
}

Node.js v20.9.0
```

## Steps to Reproduce the Problem

Shown above.

jsumners-nr:
At https://github.com/DataDog/import-in-the-middle/blob/c3c2c52c1915b47994af59d507c59029c1f1fae9/hook.js#L156-L157

[...]

Issue #62, opened 2024-01-26
Import of @react-email/components breaks with import-in-the-middle


dawnmist:
`@react-email/components` is a metapackage that re-exports the individual component libraries from `@react-email` such as `@react-email/body`, `@react-email/html`, `@react-email/tailwind`, etc.

It does this using `export * from <package>` in its index.mjs file, and the following (compiled) code for index.js:

```
module.exports = __toCommonJS(src_exports);
__reExport(src_exports, require("@react-email/body"), module.exports);
__reExport(src_exports, require("@react-email/button"), module.exports);
__reExport(src_exports, require("@react-email/column"), module.exports);
// etc for the rest of the component libraries

// Annotate the CommonJS export names for ESM import in node:
0 && (module.exports = {
  ...require("@react-email/body"),
  ...require("@react-email/button"),
  ...require("@react-email/column"),
  // etc for the rest of the component libraries
});
```

I have created an example minimal reproduction at: https://github.com/dawnmist/import-in-the-middle-react-email-issue

## Expected Behavior

The application should be able to import any of the individual components directly from the `@react-email/components` library.

## Actual Behavior

When `import-in-the-middle` is used as an `--experimental-loader`, the import paths for the `@react-email` components get incorrectly mapped as subdirectories/siblings of the index.js/index.mjs files of the `@react-email/components` library, resulting in file not found import errors:

Yarn 4 (nodeLinker=pnpm):
```sh
node:internal/process/esm_loader:40
      internalBinding('errors').triggerUncaughtException(
                                ^
[Error: ENOENT: no such file or directory, open '/home/username/project/node_modules/.store/@react-email-components-virtual-daf4c187d9/package/dist/@react-email/body'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/home/username/project/node_modules/.store/@react-email-components-virtual-daf4c187d9/package/dist/@react-email/body'
}
```

Yarn 3 (nodeLinker=pnpm):
```sh
(node:1753054) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("import-in-the-middle/hook.mjs", pathToFileURL("./"));'
(Use `node --trace-warnings ...` to show where the warning was created)

node:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^
[...]

Issue #59, opened 2024-01-25
no such file or directory


gajus:
## Expected Behavior

`getExports` is throwing an error:

```
{
  context: { format: 'module', importAttributes: {} },
  error: Error: ENOENT: no such file or directory, open '/Users/x/Developer/contra/gaia/node_modules/.pnpm/graphql-yoga@5.1.1_graphql@16.8.1/node_modules/graphql-yoga/esm/@graphql-yoga/logger'
      at async open (node:internal/fs/promises:633:25)
      at async readFile (node:internal/fs/promises:1242:14)
      at async getSource (node:internal/modules/esm/load:46:14)
      at async defaultLoad (node:internal/modules/esm/load:137:34)
      at async nextLoad (node:internal/modules/esm/hooks:750:22)
      at async getExports (/Users/x/Developer/contra/gaia/node_modules/.pnpm/import-in-the-middle@1.7.3/node_modules/import-in-the-middle/lib/get-exports.js:68:17)
      at async processModule (/Users/x/Developer/contra/gaia/node_modules/.pnpm/import-in-the-middle@1.7.3/node_modules/import-in-the-middle/hook.js:134:23)
      at async processModule (/Users/x/Developer/contra/gaia/node_modules/.pnpm/import-in-the-middle@1.7.3/node_modules/import-in-the-middle/hook.js:160:20)
      at async getSource (/Users/x/Developer/contra/gaia/node_modules/.pnpm/import-in-the-middle@1.7.3/node_modules/import-in-the-middle/hook.js:269:60)
      at async load (/Users/x/Developer/contra/gaia/node_modules/.pnpm/import-in-the-middle@1.7.3/node_modules/import-in-the-middle/hook.js:334:26) {
    errno: -2,
    code: 'ENOENT',
    syscall: 'open',
    path: '/Users/x/Developer/contra/gaia/node_modules/.pnpm/graphql-yoga@5.1.1_graphql@16.8.1/node_modules/graphql-yoga/esm/@graphql-yoga/logger'
  },
  url: 'file:///Users/x/Developer/contra/gaia/node_modules/.pnpm/graphql-yoga@5.1.1_graphql@16.8.1/node_modules/graphql-yoga/esm/@graphql-yoga/logger'
}
```

I added the following code to catch the above error.

```ts
let parentCtx;

try {
  parentCtx = await parentLoad(url, context)
} catch (error) {
  console.log({
    context,
    error,
    url
  });

  throw error;
}
```

[...]

PR #78, opened 2024-05-23
fix: Resolve re-exports of external modules

timfish:
Fixes (potentially) #59, #62, #63

`parentResolve` is cached and later used when attempting to load bare specifiers.


hook.js
```diff
@@ -91,6 +91,10 @@ function isStarExportLine (line) {
   return /^\* from /.test(line)
 }
 
+function isBareSpecifier (specifier) {
+  return /^[a-zA-Z@]/.test(specifier) && !specifier.startsWith('file:')
```
AbhiPrasad, line 95: are there are other absolute specifiers we need to watch out for here? `data:` and `node:` comes to my mind here.

Also I guess if someone is using a custom loader they can put any kind of specifier right? 😬 Like the [https one](https://nodejs.org/api/module.html#import-from-https) in the Node.js docs.

Maybe it's safer to just search for `:`?

timfish: Ah good point 🤔

Thinking about all different variants passed through here:
https://github.com/DataDog/import-in-the-middle/blob/00b01fff1f5b69dd25e307593ec54d1d8abb4844/hook.js#L157

We actually only to want to skip this code and resolve for bare specifiers. All other protocols including `file:` will be passed straight through the `URL` constructor whatever `srcUrl` and hit the next loader. The second URL constructor parameter `base` is only used if the first parameter is a relative reference and that's what trips with bare specifiers.

So we can drop the `&& !specifier.startsWith('file:')`!

timfish: No you're correct, I need to exclude all valid URLs that match the regex!

bengl (dismissed):

bengl:
@timfish eslint is failing here. Can you fix that? Other than that, LGTM.

timfish:
Should be fixed now!

bengl (approved):

PR #80, opened 2024-05-24
test: skip static-import on v22

RafaelGSS:
This test is making this module fail on v22 CITGM. This should give time to us to improve this test scenario for v23 >=.

Refs: https://github.com/nodejs/citgm/issues/1060


RafaelGSS:
> Does this not just need fixing for v22?

Possibly, yes. I just made this naive "solution" to unblock this module to run in the Node.js CITGM between releases.

timfish (changes requested):
The test still fails with this PR because the imports for `test/hook/v18-static-import-assert.mjs` are evaluated before the version check.

#81 does the version evaluation before attempting to load the test file

PR #87, opened 2024-05-29
fix: Incorrect exports

### LINKED ISSUE #60: Identifier '$Completions' has already been declared

### LINKED ISSUE #62: Import of @react-email/components breaks with import-in-the-middle

### LINKED ISSUE #68: `renamedExport` handling incorrectly adds ES module exports

### LINKED ISSUE #77: Does not handle CallExpression in ExportDefaultDeclaration node

timfish:
Closes #68, Closes #77, Closes #62, Closes #60...

We need to mimic the behaviour of Node ESM resolution and my testing so far **without** the loader suggests that:
- `export * from` only exports named exports and we don't need to pass through default exports
- If there are duplicate named exports, none of these should be exported

Ideally we should have tests that ensure that module exports are identical both with and without the loader hook because currently this isn't the case. 

To accomplish the above:
- I've added `mapExcludingDuplicates` for `setters` which does as it suggests
- Removed auto-renaming of default exports as this doesn't appear to be what Node does (ie. #68)
- Use the export name as the Map key so we remove duplicates
- In the emitted code, only add exports to `_` if there a corresponding key in `set`. Without this, the exports passed to `Hook` don't match the final module exports

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -124,7 +124,7 @@ function mapExcludingDuplicates () {
       if (map.has(k)) {
         duplicates.add(k)
         map.delete(k)
-      } else {
+      } else if (!duplicates.has(k)) {
         map.set(k, v)
       }
     },
--- a/hook.js
+++ b/hook.js
@@ -116,6 +116,24 @@ function isBareSpecifier (specifier) {
   }
 }
 
+function mapExcludingDuplicates () {
+  const map = new Map()
+  const duplicates = new Set()
+  return {
+    set (k, v) {
+      if (map.has(k)) {
+        duplicates.add(k)
+        map.delete(k)
+      } else {
+        map.set(k, v)
+      }
+    },
+    values: () => map.values(),
+    entries: () => map.entries()
+  }
+}
+
+
 /**
  * @typedef {object} ProcessedModule
  * @property {string[]} imports A set of ESM import lines to be added to the
@@ -169,12 +187,7 @@ async function processModule ({
   // As we iterate found module exports we will add setter code blocks
   // to this map that will eventually be inserted into the shim module's
   // source code. We utilize a map in order to prevent duplicate exports.
-  // As a consequence of default renaming, it is possible that a file named
-  // `foo.mjs` which has `export function foo() {}` and `export default foo`
-  // exports will result in the "foo" export being defined twice in our shim.
-  // The map allows us to avoid this situation at the cost of losing the
-  // named export in favor of the default export.
-  const setters = new Map()
+  const setters = mapExcludingDuplicates()
 
   for (const n of exportNames) {
     if (isStarExportLine(n) === true) {
@@ -208,25 +221,7 @@ async function processModule ({
       continue
     }
 
-    const matches = /^rename (.+) as (.+)$/.exec(n)
-    if (matches !== null) {
-      // Transitive modules that export a default identifier need to have
-      // that identifier renamed to the name of module. And our shim setter
-      // needs to utilize that new name while being initialized from the
-      // corresponding origin namespace.
-      const renamedExport = matches[2]
-      setters.set(`$${renamedExport}${ns}`, `
-      let $${renamedExport} = ${ns}.default
-      export { $${renamedExport} as ${renamedExport} }
-      set.${renamedExport} = (v) => {
-        $${renamedExport} = v
-        return true
-      }
-      `)
-      continue
-    }
-
-    setters.set(`$${n}` + ns, `
+    setters.set(n, `
     let $${n} = ${ns}.${n}
     export { $${n} as ${n} }
     set.${n} = (v) => {
@@ -320,30 +315,11 @@ function createHook (meta) {
       })
       const setters = Array.from(mapSetters.values())
 
-      // When we encounter modules that re-export all identifiers from other
-      // modules, it is possible that the transitive modules export a default
-      // identifier. Due to us having to merge all transitive modules into a
-      // single common namespace, we need to recognize these default exports
-      // and remap them to a name based on the module name. This prevents us
-      // from overriding the top-level module's (the one actually being imported
-      // by some source code) default export when we merge the namespaces.
-      const renamedDefaults = setters
-        .map(s => {
-          const matches = /let \$(.+) = (\$.+)\.default/.exec(s)
-          if (matches === null) return undefined
-          return `_['${matches[1]}'] = ${matches[2]}.default`
-        })
-        .filter(s => s)
-
       // The for loops are how we merge namespaces into a common namespace that
       // can be proxied. We can't use a simple `Object.assign` style merging
       // because transitive modules can export a default identifier that would
       // override the desired default identifier. So we need to do manual
       // merging with some logic around default identifiers.
-      //
-      // Additionally, we need to make sure any renamed default exports in
-      // transitive dependencies are added to the common namespace. This is
-      // accomplished through the `renamedDefaults` array.
       return {
         source: `
 import { register } from '${iitmURL}'
@@ -358,15 +334,15 @@ const primary = namespaces.shift()
 for (const [k, v] of Object.entries(primary)) {
   _[k] = v
 }
+
+${setters.join('\n')}
+
 for (const ns of namespaces) {
   for (const [k, v] of Object.entries(ns)) {
     if (k === 'default') continue
-    _[k] = v
+    if (k in set) _[k] = v
   }
 }
-
-${setters.join('\n')}
-${renamedDefaults.join('\n')}
 
 register(${JSON.stringify(realUrl)}, _, set, ${JSON.stringify(specifiers.get(realUrl))})
 `
--- a/test/hook/v18.19-static-import-gotalike.mjs
+++ b/test/hook/v18.19-static-import-gotalike.mjs
@@ -3,11 +3,6 @@ import Hook from '../../index.js'
 Hook((exports, name) => {
   if (/got-alike\.mjs/.test(name) === false) return
 
-  const bar = exports.something
-  exports.something = function barWrapped () {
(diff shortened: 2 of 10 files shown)


test/hook/v18.19-static-import-gotalike.mjs
jsumners-nr, line 1: The intention of this test is to verify that we can load modules like `got`. Can you clarify the reasoning behind removal of the removed code?

hook.js
```diff
@@ -116,6 +116,24 @@ function isBareSpecifier (specifier) {
...
 
+function mapExcludingDuplicates () {
+  const map = new Map()
+  const duplicates = new Set()
+  return {
+    set (k, v) {
+      if (map.has(k)) {
+        duplicates.add(k)
+        map.delete(k)
+      } else {
```
mohd-akram, line 127: I think this might have to be a `else if (!duplicates.has(k))`, otherwise on the next run it can get added again.

test/hook/v18.19-static-import-gotalike.mjs
timfish: It's adding exports that don't exist when the loader is not in use. When testing Node without the loader, only the root default export is exposed. `export * from` does not include default exports.

We probably just need to add a load of troublesome libraries to `devDependencies` and actually test them in the tests rather than rely on synthetic test cases.

jsumners-nr: > We probably just need to add a load of troublesome libraries to `devDependencies` and actually test them in the tests rather than rely on synthetic test cases.

Agreed.

jsumners-nr (approved):
Looks good to me.

timfish:
Closing in favour of #85 which removes more redundant code

PR #88, opened 2024-05-29
feat: Add `Hook` named export

timfish:
From [here](https://github.com/open-telemetry/opentelemetry-js/pull/4546#issue-2186767832):

> Specifically, namespace imports are required by the ESM spec to be objects. IITM's top-level export is not an object, but a callable function

Causes [issues](https://github.com/open-telemetry/opentelemetry-js/issues/4717) with some bundlers. Obviously bundled imports will not go through `import-in-the-middle` but using it should not inhibit bundlers as iitm can still hook build-in modules.

This PR adds a `Hook` named export and changes one cjs and esm test to use that new export.

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,5 +89,6 @@ Hook.prototype.unhook = function () {
 }
 
 module.exports = Hook
+module.exports.Hook = Hook
 module.exports.addHook = addHook
 module.exports.removeHook = removeHook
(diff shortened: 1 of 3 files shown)


trentm (approved):

bengl:
@timfish should there best test additions too? I.e. if something's exported, it should be tested that that interface works?

timfish:
> @timfish should there best test additions too? I.e. if something's exported, it should be tested that that interface works?

I changed both an ESM and CJS test to use this new export!

bengl (approved):

PR #93, opened 2024-06-02
test: Test 250+ popular npm packages

timfish:
Adds a script that tests 250+ of the most popular npm packages. It compares the exports without and with the loader hook and fails if they don't match. It also fails if errors are thrown with the loader hook in use.


test/check-exports/test.mjs
```diff
@@ -0,0 +1,109 @@
...
+
+let errored = false
+
+function getExports (name, loader) {
+  const args = ['--input-type=module', '--no-warnings', '-e', `import * as lib from '${name}'; console.log(JSON.stringify(Object.keys(lib)))`]
+  if (loader) args.push(loader)
+  const out = spawnSync(process.execPath, args, { cwd })
+  if (out.status !== 0) {
+    console.error(out.stderr.toString())
+    throw new Error()
```
AbhiPrasad, line 17: we should add an error message here given we are logging it in `console.error` below. 

What do you think about adding `out.error` as `error.cause` of this thrown error? That allows us to get more details about why the child process failed.

test/check-exports/test.mjs
```diff
@@ -0,0 +1,109 @@
...
+  '@babel/helper-function-name', 'p-map', 'postcss-value-parser', 'indent-string',
+  '@babel/helper-module-transforms', 'object-assign', 'delayed-stream',
+  '@nodelib/fs.stat', 'require-directory', 'diff', 'parse5',
+  'asynckit', 'tmp', 'combined-stream']
+
+console.log(`📦  Installing ${modules.length} libraries...`)
+installLibs(modules)
+
+for (const mod of modules) {
+  testLib(mod)
```
AbhiPrasad, line 103: I'd rather if we returned `errored` from `testLib` and mutate it there, and then move the `let errored` declaration down here.

Makes the side effect easier to understand (and you could even do a `array.map().some()` call too).

test/check-exports/test.mjs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function testLib (name) {
+  try {
+    const expected = getExports(name)
+    const actual = getExports(name, `--experimental-loader=${hook}`)
+    deepStrictEqual(actual, expected, `Exports for ${name} are different`)
+    console.log(`✅  Exports for '${name}' match`)
+  } catch (err) {
+    console.error(`❌  Error getting exports for '${name}':`, err)
```
AbhiPrasad, line 30: For debugging it would be nice to get the version of the package installed. What do you think about logging out the version via `npm list`?

```
spawnSync('npm', ['list', name], { cwd })
```

test/check-exports/test.mjs
```diff
@@ -0,0 +1,109 @@
...
+    errored = true
+  }
+}
+
+function installLibs (names) {
+  spawnSync('npm', ['init', '-y'], { cwd })
+  spawnSync('npm', ['install', ...names], { cwd })
+}
+
+const modules = [
```
AbhiPrasad, line 40: nit: could we leave a comment about how these are the most popular libs, and how someone can update them?

AbhiPrasad (commented):
Nice!

We're not adding this to CI yet? We can just comment out the packages that don't work for now.

Aside from these popular npm packages, could we also test against some of the popular web frameworks? `express`, `fastify`, `hapi`, `connect`, `sveltekit`, `nextjs`, `gatsby`, `remix`. These are the most common places where users will use APM instrumentation, so I think it's important we catch any issues asap.

test/check-exports/test.mjs
timfish: Yeah, can switch it around. The main thing I wanted was for the tests to continue on failures

timfish:
> We're not adding this to CI yet? We can just comment out the packages that don't work for now.

This can be added to CI, but likely not for every Node version that get tested because some of these libs won't support such a wide range. These tests with then potentially break often as the versions aren't pinned and they will update their supported node versions.

Maybe it's better the check in the `package.json` to make them more reliable?
 
> could we also test against some of the popular web frameworks? `express`, `fastify`, `hapi`, `connect`, `sveltekit`, `nextjs`, `gatsby`, `remix`. These are the most common places where users will use APM instrumentation, so I think it's important we catch any issues asap.

Yeah it's certainly worth adding more to this list. Initially this was to find potential gaps in the existing unit tests.

If we pin the versions in a `package.json`, we will not get new failures when libraries break.

test/check-exports/test.mjs
bengl: @timfish In addition to all the comments from AbhiPrasad, could you do this list as one-per-line? I'd imagine the list will change over time, and that will make the diffs easier.

bengl (dismissed):

timfish:
I've now added this to CI testing and you can see it running [in my fork](https://github.com/timfish/import-in-the-middle/actions/runs/9548533823/job/26315926666).

test/check-exports/test.mjs
```diff
@@ -0,0 +1,312 @@
+import { spawnSync } from 'child_process'
+import { deepStrictEqual } from 'assert'
+import { dirname, resolve } from 'path'
+import { fileURLToPath } from 'url'
+
+const cwd = dirname(fileURLToPath(import.meta.url))
+const hook = resolve(cwd, '..', '..', 'hook.mjs')
+
+const modules = [
```
AbhiPrasad, line 9: let's add a comment that this is the 250+ most popular packages on npm (and maybe even extract the web frameworks out to be a separate array for better readability).

AbhiPrasad (approved):

Qard (approved):

.github/workflows/ci.yml
```diff
@@ -103,3 +103,15 @@ jobs:
...
+
+  integration:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # We only test that exports match on a single node version because many of
+      # the libraries we test do not support older node versions
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
```
jsumners-nr, line 115:
```suggestion
          node-version: 'lts'
```
```suggestion
          node-version: '22.x'
```

jsumners-nr (approved):

.github/workflows/ci.yml
timfish: It looks like change actually fails on CI:
https://github.com/nodejs/import-in-the-middle/actions/runs/9660682073/job/26827816770?pr=93

timfish: Turns out we need `lts/*`

jsumners-nr (approved):

PR #96, opened 2024-06-04
fix: CommonJs bare specifier resolution

### LINKED ISSUE #95: `getFullCjsExports` does not resolve re-exports from external dependencies
This had been fixed for ESM but not for CJS.

The following code should result in a default export
`test.cjs`
```ts
module.exports = require('util').deprecate;
```

Instead it results in the following error:
```
                                ^
Error: ENOENT: no such file or directory, open '/Users/tim/Documents/Repositories/import-in-the-middle/test/check-exports/util'
    at Object.readFileUtf8 (node:internal/fs/sync:25:18)
    at Object.readFileSync (node:fs:441:19)
    at getExports (/Users/tim/Documents/Repositories/import-in-the-middle/lib/get-exports.js:69:17)
    at async Promise.all (index 0)
    at async getFullCjsExports (/Users/tim/Documents/Repositories/import-in-the-middle/lib/get-exports.js:23:9)
    at async processModule (/Users/tim/Documents/Repositories/import-in-the-middle/hook.js:131:23)
    at async getSource (/Users/tim/Documents/Repositories/import-in-the-middle/hook.js:249:23)
    at async load (/Users/tim/Documents/Repositories/import-in-the-middle/hook.js:280:26)
    at async nextLoad (node:internal/modules/esm/hooks:833:22)
    at async Hooks.load (node:internal/modules/esm/hooks:416:20) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/tim/Documents/Repositories/import-in-the-middle/test/check-exports/util'
}
```

I think this is due to this code which doesn't support resolving bare specifiers or node built-ins?
https://github.com/DataDog/import-in-the-middle/blob/a8da141be1e0fe802cea9bdb83d39a61c3aea7f7/lib/get-exports.js#L20-L30

timfish:
Closes #95

I've also made the `getExports` functions return `Set<string>` instead of `string[]` as this saves a load of unnecessary allocations. 

I've also run this through the additional module tests in #93 and get the same results.

### DIFF
--- a/lib/get-exports.js
+++ b/lib/get-exports.js
@@ -1,36 +1,46 @@
 'use strict'
 
 const getEsmExports = require('./get-esm-exports.js')
-const { parse: getCjsExports } = require('cjs-module-lexer')
-const fs = require('fs')
+const { parse: parseCjs } = require('cjs-module-lexer')
+const { readFileSync } = require('fs')
+const { builtinModules } = require('module')
 const { fileURLToPath, pathToFileURL } = require('url')
+const { dirname } = require('path')
 
 function addDefault (arr) {
-  return Array.from(new Set(['default', ...arr]))
+  return new Set(['default', ...arr])
 }
 
 const urlsBeingProcessed = new Set() // Guard against circular imports.
 
-async function getFullCjsExports (url, context, parentLoad, source) {
+async function getCjsExports (url, context, parentLoad, source) {
   if (urlsBeingProcessed.has(url)) {
     return []
   }
   urlsBeingProcessed.add(url)
 
-  const ex = getCjsExports(source)
-  const full = Array.from(new Set([
-    ...addDefault(ex.exports),
-    ...(await Promise.all(ex.reexports.map(re => getExports(
-      (/^(..?($|\/|\\))/).test(re)
-        ? pathToFileURL(require.resolve(fileURLToPath(new URL(re, url)))).toString()
-        : pathToFileURL(require.resolve(re)).toString(),
-      context,
-      parentLoad
-    )))).flat()
-  ]))
+  try {
+    const result = parseCjs(source)
+    const full = addDefault(result.exports)
 
-  urlsBeingProcessed.delete(url)
-  return full
+    for (const re of result.reexports) {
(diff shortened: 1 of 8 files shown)


lib/get-exports.js
```diff
@@ -1,36 +1,46 @@
...
-      parentLoad
-    )))).flat()
-  ]))
+  try {
+    const result = parseCjs(source)
+    const full = addDefault(result.exports)
 
-  urlsBeingProcessed.delete(url)
-  return full
+    for (const re of result.reexports) {
```
AbhiPrasad, line 26: I think we still should `Promise.all` because of the `await getExports` call.

lib/get-exports.js
```diff
@@ -57,23 +67,23 @@ async function getExports (url, context, parentLoad) {
...
-    return addDefault(Object.keys(require(url)))
-  }
-
   if (!source) {
-    // Sometimes source is retrieved by parentLoad, sometimes it isn't.
-    source = fs.readFileSync(fileURLToPath(url), 'utf8')
+    if (format === 'builtin') {
+      // Builtins don't give us the source property, so we're stuck
+      // just requiring it to get the exports.
+      return addDefault(Object.keys(require(url)))
```
AbhiPrasad, line 74: should we cache this? I guess should we cache the `fs.readFileSync` as well?

AbhiPrasad (approved):
seems reasonable to me!

lib/get-exports.js
timfish: Possibly, but this is only for built-ins. I'll benchmark it to see how slow it is.

I don't think we should cache the `readFileSync` calls because most files will only be hit once and we'll be left hanging onto the memory.

timfish: In all the tests this specific require is only hit once so it's quite a niche code path.

There's another `Object.keys(require(id))` further down which is hit for every build-in module load from CJS code and the maximum I see is for `node:child_process` at about 1ms. These are probably worth caching because there's a limited number of possible built-ins.

bengl (approved):

PR #104, opened 2024-06-12
fix: Fallback to `parentLoad` if parsing fails

### LINKED ISSUE #101: Should fail gracefully when exports cannot be parsed
If exports cannot be parsed from source, iitm should fail gracefully and fall back to not wrapping modules.

iitm can fail to parse the source for a number of reasons:
- Another loader is in use and is loading non-javascript files
- https://github.com/getsentry/sentry-javascript/issues/12422

timfish:
Closes #101


Qard (approved):

AbhiPrasad (approved):

timfish:
I added a test which shows how this fixes #105

timfish:
The tests ran [on my own fork](https://github.com/timfish/import-in-the-middle/actions/runs/9489242307/job/26150076568) but show this failing on node v16 due to the expected rejection in `import-executable.mjs` but looking at it, I would expect this to fail on every Node version! 
I guess it's an issue with top-level-await and `assert.rejects` across node versions?

bengl (approved):

PR #112, opened 2024-06-19
feat: Add environment variable to disable warnings

timfish:
This PR adds a `DISABLE_IITM_WARNINGS` environment variable that can be set to disable the logging of warnings when we fail to wrap a module.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -290,20 +290,22 @@ register(${JSON.stringify(realUrl)}, _, set, ${JSON.stringify(specifiers.get(rea
 `
         }
       } catch (cause) {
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

jsumners-nr (commented):
I think it would be better to use Node's built-in mechanism for this. See https://github.com/fastify/process-warning for a module that makes it easy (really just port the concept into here).

timfish:
> I think it would be better to use Node's built-in mechanism for this

Yep, seems like a better approach!

timfish:
Closing in favour of #114

Issue #38, opened 2023-11-22
Assignments to exported variables are not visible in other modules


dennisjlee:
## Expected Behavior

If a module exports a variable that was defined with `let` and can be reassigned, other modules that import that variable should always see the current value of the variable.

E.g.

env.js:
```
let env = { FOO: 'baz' };   // Starts with an old value

function setEnv(newEnv) {
  // Allow callers to set a new value
  env = newEnv;
}

export { env, setEnv };
```

If module1 calls `setEnv` with a new value, and module2 references `env`, module2 should see the new value.

## Actual Behavior

With the Node flag `--experimental-loader=import-in-the-middle/hook.mjs`, module2 actually sees the old value.

## Steps to Reproduce the Problem

I made a repro here: https://github.com/dennisjlee/iitm-assign-module-var-repro

## Specifications

  - Version: import-in-the-middle 1.4.2
  - Platform: Node 18.18.2 or 20.9.0

timfish:
Because the iitm wrappers assign each export to a variable before re-exporting, primitive exports have their original value copied and the reference is lost.

This is where the code generation does that:
https://github.com/nodejs/import-in-the-middle/blob/57d2459748492abd55e04a679d8b5752b90dfab4/hook.js#L189-L190

Possible solutions/workarounds? There's not much we can actually do to fix this.

For modules loaded via bare specifiers (ie. `openai` in this reproduction case #113), we could only wrap the module at it's publicly exposed exports. Currently we wrap every file individually as the modules full source files are imported.

This would not fix the issue completely but would stop it from breaking mutability internally in modules.

This is potentially a breaking change because it changes the existing behaviour. Is it important for iitm to be able to Hook libraries internally?

dennisjlee:
@timfish I encountered this issue when working with a SvelteKit application. I was trying to update our NewRelic Node.js agent to a more recent version; NewRelic switched to using `import-in-the-middle` to instrument module loading in v11 ([release notes](https://docs.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes/node-agent-11-0-0/)).

When I was testing that setup with the SvelteKit app, this turned out to be a blocking bug because of [this module in SvelteKit](https://github.com/sveltejs/kit/blob/main/packages/kit/src/runtime/shared-server.js) that relies on reassigning exported symbols, like my simplified repro above.

[...]

PR #113, opened 2024-06-20
Introduce failing test to recreate openai shim issue

ericallam:
Using iitm 1.8.1 in an ESM project that imports the `openai` package leads to the following exception:

```
TypeError: getDefaultAgent is not a function
```

`getDefaultAgent` is part of the openai [shims](https://github.com/openai/openai-node/blob/master/src/_shims/index.mjs) system, which automatically sets certain platform/runtime specific values using module side-effects. I've recreated a simplified version of this issue in a test file (I may have put the test file in the wrong place, please advise and I can move it if needed). You can see the test pass with the following:

```sh
node --require ./test/version-check.js  test/hook/side-effects.mjs
```

And fail with the loader added:

```sh
node --require ./test/version-check.js --experimental-loader ./test/generic-loader.mjs test/hook/side-effects.mjs

AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ undefined
- 'node'
    at file:///Users/eric/code/triggerdotdev/opensource/import-in-the-middle/test/hook/side-effects.mjs:4:1
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: undefined,
  expected: 'node',
  operator: 'strictEqual'
}
```

I'm not exactly sure what the issue is, but it does look like possibly the second import of `registry.mjs` in the shims file maybe just uses a module cache, instead of re-importing the module:

```
import * as shims from "./registry.mjs";
import * as auto from "./runtime.mjs";

if (!shims.kind) {
  shims.setShims(auto.getRuntime());
}

export * from "./registry.mjs";
```

### DIFF
--- a/test/fixtures/side-effects/core.mjs
+++ b/test/fixtures/side-effects/core.mjs
@@ -1,0 +1,2 @@
+import { kind as shimsKind } from "./shims.mjs";
+export { shimsKind };
--- a/test/fixtures/side-effects/registry.mjs
+++ b/test/fixtures/side-effects/registry.mjs
@@ -1,0 +1,9 @@
+export let kind = undefined;
+
+export function setShims(shims) {
+  if (kind) {
+    throw new Error(`shims already set to ${kind}`);
+  }
+
+  kind = shims.kind;
+}
--- a/test/fixtures/side-effects/runtime.mjs
+++ b/test/fixtures/side-effects/runtime.mjs
@@ -1,0 +1,5 @@
+export function getRuntime() {
+  return {
+    kind: "node",
+  };
+}
--- a/test/fixtures/side-effects/shims.mjs
+++ b/test/fixtures/side-effects/shims.mjs
@@ -1,0 +1,8 @@
+import * as shims from "./registry.mjs";
+import * as auto from "./runtime.mjs";
+
+if (!shims.kind) {
+  shims.setShims(auto.getRuntime());
+}
+
+export * from "./registry.mjs";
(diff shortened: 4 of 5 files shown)


ericallam:
Seems like other people are running into this issue in the wild, as seen in this issue here: https://github.com/openai/openai-node/issues/903

timfish:
I'm looking into this!

timfish:
@ericallam do you happen to have the openai code that causes `TypeError: getDefaultAgent is not a function`?

ericallam:
@timfish full reproduction with OpenAI SDK here: https://github.com/ericallam/iitm-openai

This is the bit of code I copied from them in the test case: https://github.com/openai/openai-node/blob/master/src/_shims/index.mjs

guidev:
Any update? this makes Datadog unusable in any node project that uses the openai sdk...

timfish:
The issue with `openai` highlights a fundamental flaw in `import-in-the-middle` (#38) that probably can't be solved with the current Node loaders API.

I have no idea how DataDog initialises `import-in-the-middle` but [this PR](https://github.com/nodejs/import-in-the-middle/pull/124) adds the ability to programatically skip wrapping of incompatible modules to work around this issue for now.

timfish:

Issue #7, opened 2021-12-02
handle non-identifier exports


bmeck:
Due to WASM and https://github.com/bmeck/proposal-arbitrary-module-namespace-identifiers JS can export names that are not identifiers, this generally only affects WASM though

See potential fix by using index rather than name like in https://github.com/nodejs/node/blob/255b1db9320df7b2e17a248ba4ab659d3f25853d/test/fixtures/es-module-loaders/mock-loader.mjs#L232

PR #115, opened 2024-06-25
fix: Skip wrapping CJS modules which result in invalid identifiers

### LINKED ISSUE #94: Doesn't handle exports with invalid identifiers
CommonJs supports exports keyed with arbitrary strings which are not valid identifiers:

`exports["unsigned short"] = "something`

This means the `webidl-conversions` package results in the following error due to invalid iitm code:
```
file:///Users/tim/Documents/Repositories/import-in-the-middle/test/check-exports/node_modules/webidl-conversions/lib/index.js?iitm=true:69
      let $unsigned short = _.unsigned short
                    ^^^^^
SyntaxError: Unexpected identifier 'short'
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:118:18)
    at callTranslator (node:internal/modules/esm/loader:273:14)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:278:30)
```

timfish:
Closes #94

Rather than trying to find a way to only export invalid identifiers in versions of Node that support exports with names like this, this PR simply skips wrapping these modules. Since modules that use exports like this are rare, it's unlikely anyone will ever want to `Hook` them and if they do we can revisit this later.

`@babel/helper-validator-identifier` was chosen to check for invalid identifiers because the code looks good, it sees 50m+ downloads per week, [it's tested](https://github.com/babel/babel/blob/cfe13c2adc3dd824fb43186acc4c85abd48c36e3/packages/babel-helper-validator-identifier/test/identifier.spec.js), it has no dependencies and maybe most importantly, it's CommonJs!


AbhiPrasad (approved):

AbhiPrasad (dismissed):

package.json
```diff
@@ -56,6 +56,7 @@
     "acorn": "^8.8.2",
     "acorn-import-attributes": "^1.9.5",
     "cjs-module-lexer": "^1.2.2",
-    "module-details-from-path": "^1.0.3"
+    "module-details-from-path": "^1.0.3",
+    "@babel/helper-validator-identifier": "^7.24.7"
```
jsumners-nr, line 60: I'd like to hear from @Qard or @bengl what they think about adding in this dependency.

jsumners-nr (changes requested):
If such modules are rare, how did the issue get exposed?

jsumners-nr:
> > If such modules are rare, how did the issue get exposed?
> 
> * When [testing the top 250+ npm packages](https://github.com/nodejs/import-in-the-middle/pull/93) I found `webidl-conversions` caused invalid wrappers to be emitted
> * [This issue](https://github.com/nodejs/import-in-the-middle/issues/7) suggests wasm can result in invalid identifiers?
> * [This open PR](https://github.com/nodejs/import-in-the-middle/pull/37) suggests at some point there was an issue with `node-machine-id`
> * Someone [reported this issue](https://github.com/getsentry/sentry-javascript/issues/12622#issue-2370560195) in the Sentry repository for `grpc-tools`
> 
> If these are the only examples we have, it suggests it's quite uncommon but its prevalent enough for it to cause issues for users.

Thank you.

MattIPv4:
👋 Is the blocker on this PR still waiting for reviews from specific folks? Anything that can be done to move that along? Happy to help where I can. This change is still required to fix a bug in Sentry that is blocking us, as I understand it.

package.json
jsumners-nr: Bump

timfish: We could vendor this code and it's around 13kB.

This package has 50m weekly downloads so there's a high chance it will already be in many users dependencies.

jsumners-nr: > This package has 50m weekly downloads so there's a high chance it will already be in many users dependencies.

Maybe. But I've never seen the package before this PR. I looked through it back when this PR was opened, and I remember getting the impression that it could be overkill. So I wanted an opinion from the folks much more in tune with this code than I am.

Qard: I'm fine with it, though I wonder if we really need _all_ that and not just a simplified subset of it.

Qard (approved):

jsumners-nr (approved):

package.json
timfish: We need all of `identifier.js` (12.2kB) which is most of the package.

image

trentm:
FWIW, it looks like this:

```
% cat test/hook/invalid-identifier.mjs
import * as lib from '../fixtures/invalid-identifier.js'
import { strictEqual } from 'assert'

strictEqual(typeof lib['one.two'], 'function')


% cat test/hook/invalid-identifier.mjs
import * as lib from '../fixtures/invalid-identifier.js'
import { strictEqual } from 'assert'

strictEqual(typeof lib['one.two'], 'function')


% cat test/fixtures/invalid-identifier.js
exports['one.two'] = () => console.log('b')


% cat my-setup-hook.mjs
import { register } from 'module'

Hook((exports, name) => {
  console.log('my-setup-hook: hooking "%s"', name)
})


% node --import ./my-setup-hook.mjs test/hook/invalid-identifier.mjs
(node:33546) Error: 'import-in-the-middle' failed to wrap 'file:///Users/trentm/tm/import-in-the-middle2/test/fixtures/invalid-identifier.js'
my-setup-hook: hooking "assert"
my-setup-hook: hooking "/Users/trentm/tm/import-in-the-middle2/test/hook/invalid-identifier.mjs"
```

With --trace-warnings:

```
% node --trace-warnings --import ./my-setup-hook.mjs test/hook/invalid-identifier.mjs
(node:33566) Warning: Error: 'import-in-the-middle' failed to wrap 'file:///Users/trentm/tm/import-in-the-middle2/test/fixtures/invalid-identifier.js'
    at load (/Users/trentm/tm/import-in-the-middle2/hook.js:431:21)
    at async nextLoad (node:internal/modules/esm/hooks:866:22)
    at async Hooks.load (node:internal/modules/esm/hooks:449:20)
    at async MessagePort.handleMessage (node:internal/modules/esm/worker:196:18) {
  cause: Error: 'one.two' export is not a valid identifier name
      at addSetter (/Users/trentm/tm/import-in-the-middle2/hook.js:205:13)
      at processModule (/Users/trentm/tm/import-in-the-middle2/hook.js:260:7)
      at async load (/Users/trentm/tm/import-in-the-middle2/hook.js:394:25)
      at async nextLoad (node:internal/modules/esm/hooks:866:22)
      at async Hooks.load (node:internal/modules/esm/hooks:449:20)
      at async MessagePort.handleMessage (node:internal/modules/esm/worker:196:18)
}
    at emitWarning (/Users/trentm/tm/import-in-the-middle2/hook.js:183:11)
    at load (/Users/trentm/tm/import-in-the-middle2/hook.js:433:9)
    at async nextLoad (node:internal/modules/esm/hooks:866:22)
    at async Hooks.load (node:internal/modules/esm/hooks:449:20)
    at async MessagePort.handleMessage (node:internal/modules/esm/worker:196:18)
my-setup-hook: hooking "assert"
my-setup-hook: hooking "/Users/trentm/tm/import-in-the-middle2/test/hook/invalid-identifier.mjs"
```

If I know about the `excludes` option (I don't recall if this is still considered experimental, hence why not documented) and with some trial and error, I can tweak `my-setup-hook.mjs` to skip that problematic module to avoid having warnings in my face all the time:

```js
...
register('./hook.mjs', import.meta.url, {
  data: {
    exclude: [/fixtures\/invalid-identifier.js$/]
  }
})
```

Then, no more warning:

```
% node --trace-warnings --import ./my-setup-hook.mjs test/hook/invalid-identifier.mjs
my-setup-hook: hooking "assert"
my-setup-hook: hooking "/Users/trentm/tm/import-in-the-middle2/test/hook/invalid-identifier.mjs"
```

trentm:
This doesn't need to block this PR, but my guess is this kind of message:

```
(node:33546) Error: 'import-in-the-middle' failed to wrap 'file:///Users/trentm/tm/import-in-the-middle2/test/fixtures/invalid-identifier.js'
```

being seen by end users will result in support issues for us. :)

That might be alleviated somewhat by getting that closer to something like (taking away the words "Error" and "failed"):

```
(node:33546) Warning: 'import-in-the-middle' could not wrap 'file:///Users/trentm/tm/import-in-the-middle2/test/fixtures/invalid-identifier.js'
```

trentm:
> That might be alleviated somewhat by getting that closer to something like (taking away the words "Error" and "failed"):

That would be possible with this diff:

```diff
diff --git a/hook.js b/hook.js
index b5d72f5..93664b8 100644
--- a/hook.js
+++ b/hook.js
@@ -428,7 +428,8 @@ register(${JSON.stringify(realUrl)}, _, set, ${JSON.stringify(specifiers.get(rea
         //
         // We log the error because there might be bugs in iitm and without this
         // it would be very tricky to debug
-        const err = new Error(`'import-in-the-middle' failed to wrap '${realUrl}'`)
+        const err = new Error(`'import-in-the-middle' could not wrap '${realUrl}'`)
+        err.name = 'Warning'
         err.cause = cause
         emitWarning(err)

```

timfish:
In the Diagnostics WorkGroup meeting we decided that this PR has enough approvals and consensus that we can merge it. 

Unfortunatly now a test fails!

timfish:
Closing in favour of #198

PR #119, opened 2024-06-28
feat: Add package.json engines to pin Node 12+

AbhiPrasad:
We should make it clear that this package only supports Node 12+. This also allows us to more easily bump the minimum Node version for future majors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
     strategy:
       matrix:
         node-version:
+          # We test against the earliest Node 12 version because that is what we
+          # support with our current engines config.
+          - 12.0.0
           - 12.x
           - 14.x
           - 16.10.0
--- a/package.json
+++ b/package.json
@@ -22,6 +22,9 @@
     "hook",
     "hooks"
   ],
+  "engines": {
+    "node": ">=12.0.0"
+  },
   "author": "Bryan English <bryan.english@datadoghq.com>",
   "license": "Apache-2.0",
   "bugs": {


bengl (changes requested):
One of the original (albeit unstated) goals of the project was to act as a normalizer for all the various versions of loader hooks and their associated breaking changes across versions of Node.js. For that reason, I wouldn't anticipate wanting to drop support for older versions of Node.js (though maybe that may change as the collaborator space grows beyond its single-vendor origins).

Regardless, setting the engines field is findamentally a good idea. My only requested change is that the earliest version claimed to be supported via the engines field ought to be also in the test matrix in CI.

AbhiPrasad:
I picked `12.x` because this what was in the matrix: https://github.com/nodejs/import-in-the-middle/blob/537cbad6de0f8c186473b2f7d7092c132625e0b2/.github/workflows/ci.yml#L28

Is there another place I should be looking?

> For that reason, I wouldn't anticipate wanting to drop support for older versions of Node.js

I think this is very reasonable, and aligns with my personal ([and professional](https://develop.sentry.dev/sdk/philosophy/#compatibility-is-king)) philosophy around this stuff.

Practically speaking though ESM is quite the broken experience in general if you use any version below 18, (`import.meta.resolve` is only `20.6`+ for example 😬), so I'm a little more comfortable being aggressive with version support because I don't assume there are many Node 12, 14, 16 apps with ESM. I would even be fine to sync up completely with OpenTelemetry (they are 14+), and adopt the same node version cadence as them because I feel like they are the primary persona this package needs to support.

What's nice is that we have all the vendors here as well, so we can easily query our systems to figure out what adoption levels are like for various Node versions to understand the ramifications of dropping node support for things.

timfish:
> I picked `12.x` because this what was in the matrix

12.x means that the latest CI runs have actually been using the latest v12 release (ie. v12.22.12) rather than 12.0.0

AbhiPrasad:
I went ahead and added `12.0.0` to the matrix. I figured this would be better because it doesn't add much to CI time, and have a more liberal engines config is easier to understand for users.

jsumners-nr (commented):
Not blocking, but putting out my views:

1. The engines block does not do anything except incur maintenance cost.
2. Tracking older than currently supported core LTS versions is too expensive. If folks insist on running on unsupported versions, they should be comfortable running unsupported packages that were released alongside those Node.js versions.

Qard:
@jsumners-nr That's a nice idea, but that's not reality for many of our customers. There are many companies locked to a Node.js runtime version but able to add and upgrade individual modules within their apps. They _expect_ to have access to our new product features and so upgrade the libraries but still expect it to work with their old runtime versions.

As much as I would _like_ to be able to drop support for EOL versions, it's not really something we can do.

trentm:
My notes on which minor/patch versions of which Node.js majors we support using IITM with the Elastic APM agent, in case it helps:

```js
// Supported Node.js version range for import-in-the-middle usage.
// - v12.20.0 add "named exports for CJS via static analysis"
//   https://nodejs.org/en/blog/release/v12.20.0
// - v14.13.1 includes "named exports for CJS via static analysis" plus some
//   fixes in the .1 minor release
// - v18.1.0 fixes an issue in v18.0.0
//   https://github.com/nodejs/node/pull/42881
// - `^18.19.0 || >=20` support was added by IITM@1.7.3
//   https://github.com/DataDog/import-in-the-middle/pull/27
// - v20.2.0 fixes an issue in v20
//   I think it is https://github.com/nodejs/node/issues/47929
```

.github/workflows/ci.yml
```diff
@@ -25,6 +25,9 @@ jobs:
     strategy:
       matrix:
         node-version:
+          # We test against the earliest Node 12 version because that is what we
+          # support with our current engines config.
+          - 12.0.0
```
trentm, line 30:
```suggestion
          - 12.20.0
```

package.json
```diff
@@ -22,6 +22,9 @@
     "hook",
     "hooks"
   ],
+  "engines": {
+    "node": ">=12.0.0"
```
trentm, line 26:
```suggestion
    "node": ">=12.20.0"
```

This doesn't fully describe the supported versions. For example, this allow 14.0.0 when one needs 14.3.1 or later on the 14.x line. However, perhaps this is good enough?

The full spelling would be something like `^12.20.0 || ^14.13.1 || ^16.0.0 || ^18.1.0 || >=20.2.0`. That excludes the odd-numbered majors, if that really matters (probably not).

Issue #122, opened 2024-07-01
Codify governance


jsumners-nr:
Now that this module is handled under the Node.js umbrella org with a variety of committing members, we need to codify how the project will be governed. Some things to address (with others likely missed):

1. PR approvals and merges: how many members with commit access need to approve a PR before it is merged? In my view, we need a minimum of two, with the approvers coming from individuals that are not on the same professional team as the person submitting the PR.
2. How will releases be handled: presumably we will implement some sort of automation. How will that work? Will a release be issued after every PR merge? On a new tag being created? And until then, who will be managing releases manually?

Here's what I do know. Node.js is an OpenJSF project, and so is Fastify. Fastify's governance passes all requirements:

* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/CODE_OF_CONDUCT.md
* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/CONTRIBUTING.md
* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/GOVERNANCE.md
* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/PROJECT_CHARTER.md
* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/SECURITY.md
* https://github.com/fastify/fastify/blob/9898d089791d30623c4f48a968ab0983c1721ae8/docs/Reference/LTS.md

timfish:
Looking through the Node libs published to NPM, they seem to use `release-please`:
- https://github.com/nodejs/node-gyp/blob/main/.github/workflows/release-please.yml
- https://github.com/nodejs/node-addon-api/blob/main/.github/workflows/release-please.yml

So we'd use the [googleapis/release-please-action](https://github.com/googleapis/release-please-action) and for that we'd need a Github auth token with the correct permissions and an npm token for publishing.

We are currently missing a `CHANGELOG.md` which is a requirement.

It [creates (and then updates) release PRs](https://github.com/googleapis/release-please-action?tab=readme-ov-file#how-release-please-works) and relies on conventional commit messages:
- `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
- `feat:` which represents a new feature, and correlates to a SemVer minor.
[...]

PR #123, opened 2024-07-03
chore: Add release workflow

timfish:
Ref #122

`release-please.yml` is copied straight from here:
https://github.com/nodejs/node-addon-api/blob/main/.github/workflows/release-please.yml

For this to actually work it needs Github credentials that can create PRs that trigger other workflows. The default GITHUB_TOKEN cannot do this.
https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials

We also need an NPM token added to the secrets so it can publish.


jsumners-nr:
The tool supports changing the log type to use the GitHub API - https://github.com/googleapis/release-please/blob/main/docs/customizing.md#changelog-types

timfish:
> The tool supports changing the log type to use the GitHub API

Ah nice. Do you know where this setting goes? It's not immediately obvious!

jsumners-nr (approved):
Looks good to me.

timfish:
Who do we need to ping to get the correct tokens added to the repo secrets? If we do that before this is merged it should create a release PR!

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,31 @@
+name: release-please
+
+on:
+  push:
+    branches:
+      - main
```
anonrig, line 6: This will release a new version on every push to main. Do we really want that, or should we go for a different approach?

anonrig:
> Who do we need to ping to get the correct tokens added to the repo secrets? If we do that before this is merged it should create a release PR!

@bengl @mhdawson can you add the repo secret to land this?

.github/workflows/release-please.yml
timfish: This runs the action on every push to main but I don't think it will release every time. On every push to main it creates/updates a PR for the next release (like [this one](https://github.com/nodejs/node-addon-api/pull/1526)) and the release is only published once this PR is merged.

anonrig (approved):

mhdawson:
@legendecas any chance you could help with adding the token since you did it recently for the node-addon-api repo and can probably do it faster than me?

legendecas:
> any chance you could help with adding the token since you did it recently for the node-addon-api repo and can probably do it faster than me?

done. Added `NPM_TOKEN` repository secret granular token, expiring on July 31, 2025.

legendecas (approved):

timfish:
Perfect, I think we just need a Github token now unless someone has done that?

[The docs](https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials) suggest the default included in GHA does not have the permissions to create PRs.

legendecas:
`GITHUB_TOKEN` is automatically generated: https://docs.github.com/en/actions/security-guides/automatic-token-authentication.

I think this should be ready.

Qard (approved):

timfish:
> GITHUB_TOKEN is automatically generated

It is automatically generated, but if PRs are created with that token, workflows don't run on them.

The [`release-please` docs](https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials) cover this and I've seen this issue on other repositories.

> By default, Release Please uses the built-in GITHUB_TOKEN secret. However, all resources created by release-please (release tag or release pull request) will not trigger future GitHub actions workflows, and workflows normally triggered by release.created events will also not run. 
>
> You will want to configure a GitHub Actions secret with a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) if you want GitHub Actions CI checks to run on Release Please PRs.

Githubs [own docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) also mention this:

> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN` will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.

Although now I'm looking at [other](https://github.com/nodejs/node-addon-api/pull/1526) PRs [created](https://github.com/nodejs/node-gyp/pull/3006) by `release-please` in other Node org repos, workflows aren't running in those so maybe there's no need for workflows to run in the release PRs?

I'll merge this and it can be added later if it's a problem!

timfish:
So the [PR has been created](https://github.com/nodejs/import-in-the-middle/pull/125) but the workflow hasn't run so there's no way to merge it!

PR #132, opened 2024-07-05
chore: More `release-please` fixes

timfish:
So the docs don't specifcally mention this but I found an [issue that mentions a solution](https://github.com/googleapis/release-please-action/issues/941#issuecomment-2036412965). If you inlcude any actions parameters other than `token`, `config-file`, or `manifest-file`, the config file will not be read correctly. 

That's a couple of hours of my life I'll never get back!


AbhiPrasad (approved):
IMO there needs to be a better error message for this, that's worth an upstream GH issue.

PR #134, opened 2024-07-06
chore: Another `release-please` fix

timfish:
So it appears that the `"changelog-type": "github"` option actully only uses Github to determine the changelog output which results in an [incorrect determination of the current release version](https://github.com/nodejs/import-in-the-middle/pull/133) and it also arguably [has a worse changelog format](https://github.com/nodejs/import-in-the-middle/pull/133/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR5).


jsumners-nr (approved):
(╯°□°）╯︵ ┻━┻

Issue #141, opened 2024-07-15
Import in the middle breaks `drizzle-orm` exports


Lms24:
## Expected Behavior

Using `import-in-the-middle` works with imports from `drizzle-orm` as with other NPM packages

## Actual Behavior

Registering the `import-in-the-middle` hook causes certain top-level exports from the `drizzle-orm` package to be missing.

This was originally [reported as a bug](https://github.com/getsentry/sentry-javascript/issues/12912) in our Sentry JS SDK repo and I tracked it down to registering the IITM hook as the only thing needed to reproduce the error. 

## Steps to Reproduce the Problem

I created a minimal reproduction example in this repo: https://github.com/Lms24/iitm-bug-drizzle-or

 1. clone repo
 2. `npm install`
 3. run `node index.js`:
     - observe 112 exports from `drizzle-orm` console log
     - script terminates successfully
 4. run `node --import instrument.mjs index.js`
      - observe 85 exports form `drizzle-orm` console log (=> 27 exports missing)
      - script crashes when calling one of the missing functions (e.g. `drizzle.or()`)

## Specifications

  - Version: `import-in-the-middle@1.9.0`
  - Platform: Node 20.11.0
  - Subsystem: MacOS 14.4

Lms24:
This is likely a duplicate of https://github.com/nodejs/import-in-the-middle/issues/139.

I'm gonna leave this open until we can confirm that #140 fixes this issue as well. If it does, sorry for the duplicate!

timfish:
This is probably not a duplicate of #139 because my current PR doesn't fix this.

timfish:
Oh wow, so this issues is caused by the fact that we are finding a duplicate export named `or`. Duplicate named exports are usually not exported up though tho the root of the module although they are in this case because they come from the same srouce file.

A simpler reproduction:
`a.mjs`
```ts
export const val = 1;
```

`b.mjs`
```ts
export * from './a.mjs';
```

`c.mjs`
```ts
export * from './a.mjs';
export * from './b.mjs';
```
`c.mjs` should have a single export named `val`.

We are not currently checking if duplicate named exports are from the same source file and are in fact, not duplicates.

BryceMehring:
Is this planned on getting fixed? Currently right now I'm importing the CJS build of drizzle-orm instead of the ESM version to avoid this issue.

jsumners-nr:
You are welcome to submit a PR with a fix.

timfish:
[...]

PR #146, opened 2024-07-16
feat: Optionally only wrap modules hooked in `--import`

timfish:
## The Problem

`import-in-the-middle` has been designed to wrap every module to allow hooking of them after they've been imported.

However `import-in-the-middle` has a couple of fundamental issues which means it will likely remain incompatible with some libraries:
- Primative exports get copied https://github.com/nodejs/import-in-the-middle/issues/38
- We can't mimic duplcate exports correctly https://github.com/nodejs/import-in-the-middle/issues/141#issuecomment-2229628535

These issues almost certainly cannot be solved with our current wrapping module strategy and are unlikely to get fixed even with the new loader proposals [without changes to the ESM spec](https://github.com/nodejs/loaders/pull/198#discussion_r1672072473). 

## A Solution?

The obvious workaround is to only wrap modules that we later intend to hook!

Since the `--loader` cli arg is deprecated, many APM vendors (including Sentry) now recommend that users initialise instrumentation of their code via the `--import` cli arg. For Sentry at least, we register the hook **and also** initialise all the OpenTelemetry plugins which in turn call `Hook` with the required libraries.

This PR adds the ability to only wrap modules that are hooked before they are imported. This is almost always the case when hooking via `--import`.

When registering the iitm hook, you can pass a `MessagePort` that is passed like this:
```ts
import { register } from 'module'
import { Hook, createAddHookMessageChannel } from 'import-in-the-middle'

const { registerOptions, waitForAllMessagesAcknowledged } = createAddHookMessageChannel()

register('import-in-the-middle/hook.mjs', import.meta.url, registerOptions)

Hook(['fs'], (exports, name) => {
 // instrument fs here
});

// Ensure that the loader has acknowledged all the modules before we allow execution to continue
await waitForAllMessagesAcknowledged()
```
Or more explicitly, you can pass the message port yourself:
```ts
const { addHookMessagePort, waitForAllMessagesAcknowledged } = createAddHookMessageChannel()

register('import-in-the-middle/hook.mjs', import.meta.url, { data: { addHookMessagePort, include: [] }, transferList: [addHookMessagePort] })
```

Now, if there are any calls to `Hook` with a specific list of modules, these will be added to the `include` list and only those modules will be wrapped. 

This would mean that if you initialise OpenTelemetry via `--import`, the OpenTelemetry plugins define which modules to wrap!


timfish:
@pichlermarc Can you think of any otel instrumentations where this might be problematic?

My first thought was that maybe [`opentelemetry-instrumentation-aws-lambda`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/afccd0d62a0ea81afb8f5609f3ee802c038d11c6/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L147) might be an issue but looking at the code, as long as the `LAMBDA_TASK_ROOT` environment variable is there, it should all work?

AbhiPrasad (dismissed):
We should update the README to add details about `createAddHookMessageChannel`

AbhiPrasad (dismissed):

Qard (commented):
I think there's a bit of a race condition here in that there's no way to be sure the messages have been received by the loader thread before the `--import` file completes and the entrypoint file begins.

Perhaps there could be some sort of acknowledgement message for each and there could be a promise you can await at the end of that `--import` file to wait until all acknowledgements have been received before proceeding?

timfish:
> I think there's a bit of a race condition here

And even if it's not possible to trigger the race condition now, it remains a possibility in the future!

I've added acknowledgment messages and updated the example in the top post.

timfish:
I'm getting exit code 13 from the new test on some versions in CI but it passes locally:

> 13: Unsettled Top-Level Await: `await` was used outside of a function in the top-level code, but the passed Promise never settled.

I guess you can't have the end of execution blocked simply waiting on a promise because it assumes it's never going to resolve.

index.js
```diff
@@ -31,6 +32,44 @@ function callHookFn (hookFn, namespace, name, baseDir) {
...
+    }
+  }).unref()
+
+  function waitForAllMessagesAcknowledged () {
+    // This timer is to prevent the process from exiting with code 13:
+    // 13: Unsettled Top-Level Await.
+    const timer = setInterval(() => { }, 1000)
+    const promise = new Promise((resolve) => {
+      resolveFn = resolve
+    }).then(() => { clearInterval(timer) })
```
timfish, line 61: This timer stops the process exiting with error code 13 which is triggered by top-level await if there are no more active handles

AbhiPrasad: I spent little bit trying to avoid this but couldn't come up with a better way. I'm gonna take another look tomorrow morning, but if I can't figure anything else out I'll approve.

jsumners-nr (dismissed):
Personally, I have a hard time following the API (e.g. I don't understand why the `options` has to be constructed by the user), but if it works 👍

AbhiPrasad (approved):

timfish:
>  I don't understand why the options has to be constructed by the user

Yes, good point. 

I've added `registerOptions` to the return of `createAddHookMessageChannel()` and retained `addHookMessagePort` in case anyone wants to be more explicit.

I've updated the above code example, the types, and the readme!

index.js
```diff
@@ -31,6 +32,47 @@ function callHookFn (hookFn, namespace, name, baseDir) {
   }
 }
 
+let sendModulesToLoader
+
+function createAddHookMessageChannel () {
```
jsumners-nr, line 37: I think that when we add new functions we should also add doc blocks for them.

jsumners-nr (dismissed):
Looks good to me.

AbhiPrasad (dismissed):

trentm:
I think I found a bug where `node:`-prefixed imports are not included when the unprefixed name is hooked:

## the control case

First, this shows hooking `os.type()`, **not** using the functionality in this PR.

`dosomething.mjs`
```js
// import * as os from 'os' // WORKS
import * as os from 'node:os' // FAILS with addHook message port
import * as fs from 'fs'
console.log('dosomething: os.type()=%s, fs.readFileSync.name=%o', os.type(), fs.readFileSync.name)
```

`mysetup-simple.mjs`
```js
import { register } from 'module'
Hook(['os'], (exports, name) => {
  console.log('mysetup-simple: hooking "%s"', name)
  if (name === 'os') {
    const origType = exports.type
    exports.type = function wrappedType () {
      return origType.call(this) + '-hooked'
    }
  }
})
```

Running that shows `Darwin-hooked`, i.e. `os.type()` was hooked:

```
% node --import ./mysetup-simple.mjs dosomething.mjs
mysetup-simple: hooking "os"
dosomething: os.type()=Darwin-hooked, fs.readFileSync.name='readFileSync'
```

## failing case

Now using this PR, with some "XXX ..." debug prints added to the IITM code.

`mysetup.mjs`
```js
import { register } from 'module'
import { Hook, createAddHookMessageChannel } from './index.js'
const { registerOptions, waitForAllMessagesAcknowledged } = createAddHookMessageChannel()

register('./hook.mjs', import.meta.url, registerOptions)

Hook(['os'], (exports, name) => {
  console.log('mysetup: hooking "%s"', name)
  if (name === 'os') {
    const origType = exports.type
    exports.type = function wrappedType () {
      return origType.call(this) + '-hooked'
    }
  }
})

await waitForAllMessagesAcknowledged()
```

Running that:

```
% node --import ./mysetup.mjs dosomething.mjs
XXX loader: includeModules=["os"]
XXX loader: skip out, "file:///Users/trentm/tm/import-in-the-middle2/dosomething.mjs" not in includeModules
XXX loader: skip out, "node:os" not in includeModules
XXX loader: skip out, "fs" not in includeModules
dosomething: os.type()=Darwin, fs.readFileSync.name='readFileSync'
```

**The relevant debug print is: `XXX loader: skip out, "node:os" not in includeModules`.**
If I change "dosomething.mjs" to use an unprefixed import -- `import * as os from 'os'` -- then the hook works.

So, I guess this isn't an issue with *this* PR, but with the `includeModules/excludeModules` support added earlier?
I'm happy to approve this and this `node:`-prefixing difference could be handled separately.
Preferences?

trentm:
# issue with `internals: true`

I think there is also an issue with `{internals: true}` usage and `includeModules`.

(This example uses `@langchain/core`, mainly because this had been a module
used for a separate (but slightly related) issue discussed in require-in-the-middle.
You don't need to know that background for this issue, but if interested see
the second part of
https://github.com/elastic/require-in-the-middle/pull/90#issuecomment-2237715709
starting with "I'm guessing a bit here".)

## control case

Put these files in the top-dir of a git clone of import-in-the-middle.

`use-langchain-tools.mjs`
```js
import * as tools from '@langchain/core/tools'
console.log('use-langchain-tools: Imported @langchain/core/tools:', tools.tool)
```

`instr-langchain-core.mjs`
```js
import { register } from 'module'
// import { Hook } from 'import-in-the-middle'
import { Hook, createAddHookMessageChannel } from './index.js'
const { registerOptions, waitForAllMessagesAcknowledged } = createAddHookMessageChannel()

const LIMIT_TO_HOOKED = process.env.LIMIT_TO_HOOKED

if (LIMIT_TO_HOOKED) {
  register('./hook.mjs', import.meta.url, registerOptions)
} else {
}

Hook(['@langchain/core'], { internals: true }, (mod, name, baseDir) => {
  if (name === '@langchain/core/dist/callbacks/manager.js') {
    console.log('Hooked name=%j: mod.CallbackManager:', name, mod.CallbackManager)
  }
  return mod
})

if (LIMIT_TO_HOOKED) {
  await waitForAllMessagesAcknowledged()
}
```

This Hook uses `internals: true` to be able to get at internal files in the `@langchain/core` package.
In the control case (without using `includeModules` functionality) this works:

```
% npm install --no-save @langchain/core
...

% node --import ./instr-langchain-core.mjs use-langchain-tools.mjs
Hooked name="@langchain/core/dist/callbacks/manager.js": mod.CallbackManager: [class CallbackManager extends BaseCallbackManager]
use-langchain-tools: Imported @langchain/core/tools: [Function: tool]
```


## failing case

When I use `includeModules` (via the functionality in this PR), it no longer hooks:

```
% LIMIT_TO_HOOKED=true node --import ./instr-langchain-core.mjs use-langchain-tools.mjs
XXX loader: includeModules=["@langchain/core"]
XXX loader: skip out, "file:///Users/trentm/tm/import-in-the-middle2/use-langchain-tools.mjs" not in includeModules
XXX loader: skip out, "@langchain/core/tools" not in includeModules
XXX loader: skip out, "./dist/tools/index.js" not in includeModules
XXX loader: skip out, "zod" not in includeModules
XXX loader: skip out, "../callbacks/manager.js" not in includeModules
... elided about 200 lines ...
use-langchain-tools: Imported @langchain/core/tools: [Function: tool]
...
XXX loader: skip out, "uuid" not in includeModules
```

This one is a bit more tricky.

Options:
1. Just document that `includeModules` (and the related feature in this PR) just doesn't work with `internals: true` and move on.
2. Consider wrapping every submodule under the hooked module names. So for example, IITM would resolve that the `"../callbacks/manager.js"` import reference is inside `@langchain/core`, which is on `includeModules`, so it would be wrapped.
3. Consider allowing specific submodule paths as **hook args**, e.g. `Hook(['@langchain/core/dist/callbacks/manager.js'], ...)`. (RITM supports this without requiring `internals: true`.)  Then the `includeModules` handling could wrap just that submodule of `@langchain/core`, without wrapping every other submodule in the package.  Considering this would definitely be a separate issue (i.e. not part of this PR).

timfish:
@trentm thanks for the feedback!

> I think I found a bug where `node:`-prefixed imports are not included when the unprefixed name is hooked

Yes, this is a bug in the previously added feature. I'll submit an additional PR to fix and test for this.

> Consider wrapping every submodule under the hooked module names

For `exclude` at least, only excluding a single module feels like a bug from a users perspective. It currently only excludes wrapping the module entry point. If a library is problematic internally, users have to resort to using a regex to actually exclude an entire library with its sub-modules.

On the other hand, `include` only wrapping a single module feels advantageous/correct if you ignore `internals: true`.

I'll take another look at the code and how `internals: true` works and get back to you!

timfish:
Ok, so I've marked this new API as experimental and this and the previous feature addition as incompatible with the `{internals: true}` Hook option. 

> Just document... doesn't work with `internals: true` and move on.

We may be able to fix this in the future!

index.d.ts
```diff
@@ -84,3 +84,39 @@ export declare function addHook(hookFn: HookFunction): void
...
 export declare function removeHook(hookFn: HookFunction): void
+
+type CreateAddHookMessageChannelReturn<Data> = {
+  addHookMessagePort: MessagePort,
+  waitForAllMessagesAcknowledged: Promise<void>
+  registerOptions: { data?: Data; transferList?: any[]; }
+}
+
+/**
+ * EXPERIMENTAL
```
trentm, line 95: Too bad TypeScript jsdoc tag support doesn't support `@experimental`. Hopefully at some point (https://github.com/microsoft/TypeScript/issues/56808)

trentm (approved):
Thanks for doing this!

timfish:
Sorry, I had to make some changes after merging in the `node:` fix PR so I'll need approvals again!

hook.js
```diff
@@ -281,13 +281,31 @@ function createHook (meta) {
...
+      if (data.addHookMessagePort) {
+        data.addHookMessagePort.on('message', (modules) => {
+          if (includeModules === undefined) {
+            includeModules = []
+          }
+
+          for (const each of modules) {
+            if (!each.startsWith('node:') && builtinModules.includes(each)) {
+              includeModules.push(`node:${each}`)
+            }
```
timfish, line 294: This is the addition I need to make!

AbhiPrasad (approved):

trentm (approved):

PR #148, opened 2024-07-21
feat: Allow regex for `include` and `exclude` options

timfish:
This PR allows you to pass regular expressions in the `include` and `exclude` options.

cc @FabianFrank and @vchirikov


AbhiPrasad (commented):
we need to update readme!

AbhiPrasad (approved):

Qard (approved):

PR #149, opened 2024-07-27
fix: `node:` prefixed build-in modules with `include`/`exclude`

timfish:
@trentm found a bug in `node:` prefixed built-in modules when using `include` and `exclude`:
https://github.com/nodejs/import-in-the-middle/pull/146#issuecomment-2253093908

This PR fixes that bug and adds some tests for it.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -4,6 +4,7 @@
 
 const { URL } = require('url')
 const { inspect } = require('util')
+const { builtinModules } = require('module')
 const specifiers = new Map()
 const isWin = process.platform === 'win32'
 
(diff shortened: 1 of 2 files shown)


hook.js
```diff
@@ -149,6 +151,14 @@ function ensureArrayWithBareSpecifiersFileUrlsAndRegex (array, type) {
     throw new Error(`'${type}' option only supports bare specifiers, file URLs or regular expressions. Invalid entries: ${inspect(invalid)}`)
   }
 
+  // Rather than evaluate whether we have a node: scoped built-in-module for
+  // every call to resolve, we just add them to include/exclude now.
+  for (const each of array) {
+    if (typeof each === 'string' && !each.startsWith('node:') && builtinModules.includes(each)) {
```
AbhiPrasad, line 157: Let's update the docstring of `ensureArrayWithBareSpecifiersFileUrlsAndRegex` to indicate that we handle this `node:` prefixed case.

AbhiPrasad (dismissed):

trentm (approved):

PR #153, opened 2024-09-26
fix: Support Hooking multiple times

### LINKED ISSUE #98: Are sequential patched supported?
As far as i understand and according to some testing being done if there are multiple multiple hooks attached to the same package the last one will be eventually used in application. Am i right? If so are there any methods to apply sequential patches to the package. The case is that i'm using library that's patched the package i want to patch.

Library version: 1.8.0
Node version: 20.11.1

timfish:
Closes #98
Also ref - https://github.com/getsentry/sentry-javascript/pull/13763

Proxies `get` so that multiple hooks do not overwrite each other.


.eslintrc.yaml
```diff
@@ -21,3 +21,6 @@ ignorePatterns:
   - test/fixtures/circular-a.js
   - test/fixtures/circular-b.js
   - test/fixtures/reexport.js
+  - test/fixtures/duplicate-explicit.mjs
+  - test/fixtures/duplicate.mjs
+  - test/fixtures/export-types/default-call-expression-renamed.mjs
```
jsumners-nr, line 26: I don't see these files anywhere in the PR?

timfish: Sorry, should have added an explanation to the PR. 

Because we use a caret version for eslint and no lock file, these files started failing with `Multiple exports of name 'foo'` [in CI](https://github.com/nodejs/import-in-the-middle/actions/runs/11050280282/job/30697711069#step:5:11) but passed locally until I nuked `node_modules` and installed again. 

So these failures were caused by a recent update in eslint!

jsumners-nr (approved):

AbhiPrasad (approved):

PR #160, opened 2024-10-12
fix: Correct type definition for waitForAllMessagesAcknowledged

valentin-nemcev:
Type definition doesn't match documentation and code, seems like it was a typo

PR #164, opened 2024-11-13
Fix waitForAllMessagesAcknowledged type

rChaoz:
Error when doing exactly what is says in the docs because of wrong type:

image

It's a function returning a promise, not a promise:

https://github.com/nodejs/import-in-the-middle/blob/ca88f2f59ad8872811d74273283b077160df1531/index.js#L83

This fixes that


dygabo (commented):
duplicate of https://github.com/nodejs/import-in-the-middle/pull/160

PR #165, opened 2024-11-26
feat: Warn on multiple hook initialization

timfish:
We've seen cases where users have initialised the `import-in-the-middle` hook more than once. I can't really think of any cases where this is desirable and it's quite difficult to diagnose so it's probably worth displaying a warning.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -279,6 +279,12 @@ function createHook (meta) {
   let includeModules, excludeModules
 
   async function initialize (data) {
+    if (global.__import_in_the_middle_initialized__) {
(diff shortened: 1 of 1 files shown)


hook.js
```diff
@@ -279,6 +279,12 @@ function createHook (meta) {
   let includeModules, excludeModules
 
   async function initialize (data) {
+    if (global.__import_in_the_middle_initialized__) {
+      console.warn("The 'import-in-the-middle' hook has already been initialized")
```
jsumners-nr, line 283: `process.emitWarning`

PR #176, opened 2025-02-03
feat: Support import attributes 

gevalo1:
Bumps the `acorn` version to the latest which now supports import attributes

### DIFF
--- a/test/hook/v20.10-static-import-attributes.mjs
+++ b/test/hook/v20.10-static-import-attributes.mjs
@@ -2,16 +2,14 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
+import Hook from '../../index.js'
 import jsonMjs from '../fixtures/json-attributes.mjs'
 import { strictEqual } from 'assert'
 
-// Acorn does not support import attributes so an error is logged but the import
-// still works!
-//
-// Hook((exports, name) => {
-//   if (name.match(/json\.mjs/)) {
-//     exports.default.data += '-dawg'
-//   }
-// })
+Hook((exports, name) => {
+  if (name.match(/json-attributes\.mjs/)) {
+    exports.default.data += '-dawg'
+  }
+})
 
-strictEqual(jsonMjs.data, 'dog')
+strictEqual(jsonMjs.data, 'dog-dawg')
--- a/lib/get-esm-exports.js
+++ b/lib/get-esm-exports.js
@@ -1,14 +1,14 @@
 'use strict'
 
 const { Parser } = require('acorn')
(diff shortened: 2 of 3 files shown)


timfish:
- This PR changes far too many things and should be 4-5 individual PRs. For example, the eslist changes should be in their own PR.
- The Node test versions have been reduced. We are unlikely to change the supported Node versions before there has been some debate between all the stakeholders that rely on `import-in-the-middle`

gevalo1:
Yeah that makes sense. I simplified this PR to only bump `acorn` for now, that should at least fix the issue.

jsumners-nr (changes requested):
If I understand correctly, this is meant to fix an issue. So there should be a test included to prove the fix is in place and guard against regressions.

gevalo1:
Hi @mark-w-obrien

I think you may be right, I added your suggestion.

---

@jsumners-nr, I agree. There are already tests related to this functionality, but I updated the existing one to reflect my changes.

test/hook/v20.10-static-import-attributes.mjs
```diff
@@ -2,16 +2,14 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
+import Hook from '../../index.js'
 import jsonMjs from '../fixtures/json-attributes.mjs'
 import { strictEqual } from 'assert'
 
-// Acorn does not support import attributes so an error is logged but the import
-// still works!
```
gevalo1, line 9: 💡 The error in this comment is also the error I saw on my end. It's gone after the changes in this PR so I guess that covers it.

jsumners-nr:
> > I don't know why CI is failing
> 
> The integration tests install and test the latest versions of a lot of libraries. Some of them have changed and are no longer work with `import-in-the-middle`.

And they run the `test:e2e` script. Which passes locally for me.

timfish:
> Which passes locally for me.

Is there a lock file hanging around that means older versions of the libraries are getting tested locally?

jsumners-nr:
No. I do not use those things.

timfish:
> I do not use those things.

`test:e2e` ends up creating a load of gitignored files in `./test/check-exports/*` which wont exist in CI and I guess could stop newer dependencies from getting tested locally? Maybe the test should purge these first?

timfish:
I just removed those files and tested again locally with Node v20.16.0 and got a single failure for `svelte@5.19.8`

timfish:
I guess this is due to differing npm output between node versions. I will fix this but it shouldn't block this being released.

timfish (approved):

jsumners-nr (approved):

PR #180, opened 2025-02-25
test: Pin `openai` test dependency to fix CI build

timfish:
Because we don't commit the lock file, changing test dependencies can break our tests!

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "eslint-plugin-promise": "^6.1.1",
     "got": "^14.3.0",
     "imhotap": "^2.1.0",
-    "openai": "^4.47.2",
+    "openai": "4.48.1",
     "ts-node": "^10.9.1",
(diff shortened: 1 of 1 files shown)


jsumners-nr:
> Because we don't commit the lock file, changing test dependencies can break our tests!

I think this is, or should be, intentional. If a new version of a "supported" library breaks things, then this library is broken and needs to be fixed. I don't think pinning is the right solution.

timfish:
I think the change in `openai` that causes these new test failures is related to one of the other bugs which we can't fix.

jsumners-nr (approved):

mohd-akram:
#181 addresses both svelte and openai. Updated it with a test case and a description of the issue and the fix.
#182 fixes globals and mime-db

With both PRs, importing all packages in the integration CI should succeed.

timfish:
Closing in favour of #181

PR #189, opened 2025-04-29
fix: Hook internal paths like `require-in-the-middle`

### LINKED ISSUE #185: Behavior is different from `require-in-the-middle`
Hey and thanks for all the hard work,

Please note that this is related to this issue I've created in the `dd-trace-js` repository as I'm not sure who owns the responsibility for this issue: https://github.com/DataDog/dd-trace-js/issues/5479

I think `import-in-the-middle` behaves differently from `require-in-the-middle` because of this line: https://github.com/nodejs/import-in-the-middle/blob/53a33a9b07799bff815864089a0c072d223df47b/index.js#L147

I think it's an optimisation to avoid sending unnecessary files to the callback but it avoids some files that should be included.

For example, when given the `filename` `file:///<redacted>/node_modules/.pnpm/@apollo+server@4.11.3_encoding@0.1.13_graphql@16.10.0/node_modules/@apollo/server/dist/esm/ApolloServer.js`

`baseDir` will be `<redacted>/node_modules/.pnpm/@apollo+server@4.11.3_encoding@0.1.13_graphql@16.10.0/node_modules/@apollo/server`

`specifiers.get(filename)` will be `./ApolloServer.js`

So when checking `!baseDir.endsWith(specifiers.get(filename))` the computation will be `false` as you can see `./ApolloServer.js` is not present in `baseDir`.

If I remove the line, it works.

For the same intended behavior, `require-in-the-middle` will provide all the files.

Thanks in advance and have a good day.

### LINKED ISSUE #187: Impossible to instrument `@modelcontextprotocol/sdk`
The `@modelcontextprotocol/sdk` package wants you to import stuff from ` "@modelcontextprotocol/sdk/server/mcp.js"` like so:

```
import { McpServer, ResourceTemplate } from "@modelcontextprotocol/sdk/server/mcp.js";
```

This import resolves to `node_modules/@modelcontextprotocol/dist/esm/sdk/server/mcp.js`.

I cannot find a way to provide a module definition to `new Hook()` that will actually wrap the module.

It seems like we are always ending up in this branch https://github.com/nodejs/import-in-the-middle/blob/53a33a9b07799bff815864089a0c072d223df47b/index.js#L147 and bailing the wrapping.

timfish:
- Closes #185 
- Closes #187

`import-in-the-middle` differs from `require-in-the-middle` in that you can't patch library internals. Open Telemetry expects these libraries to behave in the same way so you can patch a library with the same hook configuration. 

- This is not technically breaking if `import-in-the-middle` was not behaving as intended and documented
- This could result in the `Hook` callbacks being called a lot more, for all internal files

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,8 +143,6 @@ function Hook (modules, options, hookFn) {
           if (baseDir) {
             if (internals) {
               name = name + path.sep + path.relative(baseDir, fileURLToPath(filename))
-            } else {
-              if (!baseDir.endsWith(specifiers.get(filename))) continue
             }
           }
           callHookFn(hookFn, namespace, name, baseDir)
--- a/test/hook/static-import-package-internals.mjs
+++ b/test/hook/static-import-package-internals.mjs
@@ -2,7 +2,7 @@ import Hook from '../../index.js'
 import { Report } from 'c8/index.js'
 import path from 'path'
 import { fileURLToPath } from 'url'
-import { strictEqual, notStrictEqual } from 'assert'
+import { strictEqual } from 'assert'
 
 const c8Dir = path.join(path.dirname(fileURLToPath(import.meta.url)), '..', '..', 'node_modules', 'c8')
 
@@ -12,4 +12,4 @@ Hook(['c8'], (exports, name, baseDir) => {
   exports.Report = () => 42
 })
 
-notStrictEqual(Report({}), 42)
+strictEqual(Report({}), 42)


timfish:
Pain. The test failures are due to `openai` adding some shims that you need to import first otherwise some side-effects throw errors 🤦🏻‍♂️

AbhiPrasad:
> in that you can't patch library internals

Does this mean you can only patch what is explicitly exposed via `package.json` exports?

timfish:
Closing in favour of #194

PR #191, opened 2025-05-08
fix: Don't attempt to wrap TypeScript modules

### LINKED ISSUE #186: Support for --experimental-strip-types
Creating an issue to track support here. (Great work on the library!)

Minimal repro: run the following with `node --experimental-strip-types --import ./register.cjs index.ts` with node v22.14.0 (or `node --import ./register.cjs index.ts` with node v23.11.0).

You'll get an error like:

```
(node:53480) Error: 'import-in-the-middle' failed to wrap 'file:///Users/chris/projects/personal/repro/index.ts'
file:///Users/chris/projects/personal/repro/index.ts?iitm=true:1
import type { Foo } from './foo.ts'
            ^

SyntaxError: Unexpected token '{'
    at compileSourceTextModule (node:internal/modules/esm/utils:344:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:106:18)
    at #translate (node:internal/modules/esm/loader:534:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:116:19)
```

---

`package.json`:

```json
{
  "type": "module",
  "dependencies": {
    "import-in-the-middle": "^1.13.1"
  }
}
```

---

`foo.ts`

```ts
export type Foo = { hello: string }
```

---

`index.ts`

```ts
import type { Foo } from './foo.ts'
```

---

`loader-hook.mjs`

```js
export * from 'import-in-the-middle/hook.mjs'
```

---

`register.cjs` (pretty much copied from [`dd-trace-js`](https://github.com/DataDog/dd-trace-js))

```js
const { register } = require('node:module')
const { pathToFileURL } = require('node:url')

register('./loader-hook.mjs', pathToFileURL(__filename), {
  data: { exclude: [/langsmith/, /openai\/_shims/, /openai\/resources\/chat\/completions\/messages/] }
})
```

timfish:
- Closes #186

This PR simply ignores TypeScript files and doesn't attempt to wrap them. 

This means you cannot `Hook` TypeScript files but it's considerably better than crashing.


timfish:
Ok, this PR breaks `ts-node` so it's going to be more complicated...

AbhiPrasad (dismissed):
Reading through https://nodejs.org/api/module.html#loadurl-context-nextload I also see a `wasm` format, should we also ignore those? We can do it in another PR.

timfish:
Maybe we should only opt into those that we know we can handle like `builtin`, `commonjs` and `module`?

AbhiPrasad (approved):

jsumners-nr (approved):

PR #194, opened 2025-05-23
feat: Optionally hook internal paths like `require-in-the-middle`

timfish:
As discussed in the Diagnostics Working Group meeting, this PR is similar to #189 but allows this change to be selectively enabled so it can be tested more thoroughly before becoming the default.

Sentry are happy to do this testing!


jsumners-nr (dismissed):

AbhiPrasad (commented):
This needs a note in the README

Issue #196, opened 2025-05-28
yargs-parser and cliui integration tests failing


AbhiPrasad:
https://github.com/nodejs/import-in-the-middle/actions/runs/15274631365/job/42957893111

```
file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/node_modules/yargs-parser/build/lib/index.js?iitm=true:26
      let $module.exports
                 ^

SyntaxError: Unexpected token '.'
    at compileSourceTextModule (node:internal/modules/esm/utils:344:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:105:18)
    at #translate (node:internal/modules/esm/loader:534:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27)

Node.js v22.15.0

❌  Error getting exports for yargs-parser@22.0.0: Error: Getting exports returned non-zero exit code 'yargs-parser'
    at getExports (file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/test.mjs:277:11)
    at testLib (file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/test.mjs:296:20)
    at file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/test.mjs:311:14
    at ModuleJob.run (node:internal/modules/esm/module_job:274:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)
```

```
file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/node_modules/cliui/index.mjs?iitm=true:26
      let $module.exports
                 ^

SyntaxError: Unexpected token '.'
    at compileSourceTextModule (node:internal/modules/esm/utils:344:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:105:18)
    at #translate (node:internal/modules/esm/loader:534:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27)

Node.js v22.15.0

❌  Error getting exports for cliui@9.0.1: Error: Getting exports returned non-zero exit code 'cliui'
    at getExports (file:///home/runner/work/import-in-the-middle/import-in-the-middle/test/check-exports/test.mjs:277:11)
[...]

PR #197, opened 2025-05-28
fix: Skip `module.exports` exports

AbhiPrasad:
fixes https://github.com/nodejs/import-in-the-middle/issues/196


jsumners-nr (approved):

hook.js
```diff
@@ -254,19 +257,25 @@ async function processModule ({ srcUrl, context, parentGetSource, parentResolve,
         addSetter(name, setter, true)
       }
     } else {
+      // Handle `module.exports` specially since it contains a dot which is invalid in variable names
+      // `module.exports` used by packages like `yargs-parser` to shim CJS support in ESM.
+      const variableName = n === 'module.exports' ? '$moduleExports' : `$${n}`
+      const objectKey = n === 'module.exports' ? JSON.stringify(n) : n
+      const objectKeyWithAccess = n === 'module.exports' ? `[${objectKey}]` : `.${objectKey}`
```
timfish, line 264: Can we use the object[bracket] syntax for all exports?

timfish:
#115 excludes any library with "invalid" exports from being wrapped by iitm.

I think there might be a way to use the approach in this PR to actually export these and keep the library instrument'able by iitm...

timfish:
For the intermediate variables we currently use something containing the export name which causes issues but there's nothing stopping us from using something unique like a counter `_1`, etc.

Is it only newer versions on Node that accept exporting names with strings like that? I suspect we might need a legacy output mode for older Node versions...

AbhiPrasad:
> Is it only newer versions of Node that accept exporting names with strings like that? I suspect we might need a legacy output mode for older Node versions...

I think so, the tests were breaking in node 12 and 14, so I pinned to node 16.

PR #198, opened 2025-06-11
fix: Account for invalid identifiers

AbhiPrasad:
Node 16 and above allows you to use interesting string identifiers like `"module.exports"` and `"unsigned short"` as valid export names.

ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

image

This PR accounts for that by updating our shim variable name logic, and adjusting how we do getters and setters so that we don't break when encoutering these "invalid" identifiers.

supercedes https://github.com/nodejs/import-in-the-middle/pull/115
resolves https://github.com/nodejs/import-in-the-middle/issues/94

If you can, please give this a test locally. I tried it myself on some test apps and worked fine.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -254,19 +254,22 @@ async function processModule ({ srcUrl, context, parentGetSource, parentResolve,
         addSetter(name, setter, true)
       }
     } else {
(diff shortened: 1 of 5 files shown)


timfish (approved):
This is awesome!

PR #202, opened 2025-06-26
fix: Incompatible with module.registerHooks

timokoessler:
This PR prevents that the load hook returns a nullish value for the source if a CJS module is loaded.
The reason for this is that Node.js will otherwise crash if the synchronous module customization hooks are used. See https://github.com/nodejs/node/issues/57327 for more information. The Node.js documentation also mentions that returning nullish values might no longer be supported in the future in general.

### DIFF
--- a/hook.js
+++ b/hook.js
@@ -473,7 +473,7 @@ register(${JSON.stringify(realUrl)}, _, set, get, ${JSON.stringify(specifiers.ge
     // The reason for this is that Node.js will otherwise crash if the synchronous
     // module customization hooks are used: https://github.com/nodejs/node/issues/57327
     // The Node.js documentation also mentions that returning nullish values will no longer be supported in future.
-    if (parentResult.format === 'commonjs' && (parentResult.source === null || parentResult.source === undefined)) {
+    if (parentResult.format === 'commonjs' && !parentResult.source) {
       parentResult.source = readFileSync(fileURLToPath(url), 'utf8')
     }
 


hook.js
```diff
@@ -466,7 +467,17 @@ register(${JSON.stringify(realUrl)}, _, set, get, ${JSON.stringify(specifiers.ge
...
     }
 
-    return parentLoad(url, context, parentLoad)
+    const parentResult = await parentLoad(url, context, parentLoad)
+
+    // Prevent returning a nullish value for the source if a CJS module is loaded.
+    // The reason for this is that Node.js will otherwise crash if the synchronous
+    // module customization hooks are used: https://github.com/nodejs/node/issues/57327
+    // The Node.js documentation also mentions that returning nullish values will no longer be supported in future.
+    if (parentResult.format === 'commonjs' && (parentResult.source === null || parentResult.source === undefined)) {
```
jsumners-nr, line 476: A coercive check is appropriate here:

```suggestion
    if (parentResult.format === 'commonjs' && !parentResult.source) {
```

jsumners-nr (approved):

jsumners-nr (approved):

timfish:
Happy to consider merging this if we can get the CI tests passing!

timokoessler:
Unfortunately, this fix appears to cause problems with the module cache in Node.js in some cases. I have added a comment to the issue: https://github.com/nodejs/node/issues/57327#issuecomment-3012503512.

PR #207, opened 2025-09-25
fix: `@vercel/nft` resolving `createRequire`

timfish:
This PR changed the ESM code to use `createRequire` to load the CJS hook:
- #205

`@vercel/nft` is used by some frameworks to determine which dependency files should be included in production without doing any bundling.

Unfortunately it looks like their support for `createRequire` requires that it is accessed on a variable called `module`:
- https://github.com/vercel/nft/pull/459/files#diff-f9de9e832021754e4f7e493614502e6a12c6adf1d75f666ffdf9c2b8de8d33bcR1090

### DIFF
--- a/hook.mjs
+++ b/hook.mjs
@@ -2,8 +2,9 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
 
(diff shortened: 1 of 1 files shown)


timfish:
Ok `@vercel/nft` is a mess and this on it's own doesn't fix it 😭

PR #210, opened 2025-09-25
fix: convert all modules running in loader thread to ESM

bengl:
This is a second attempt at fixing the double-loader issue in #205. The tests from that commit are included.


timfish (approved):
Looks good to me!

@bengl did you want to make this a new semver major?

timfish:
We should probably merge #211 before we cut a major!

Issue #90, opened 2024-05-30
Use `es-module-lexer` to parse ESM


timfish:
@AbhiPrasad suggested we use `es-module-lexer` father than `acorn`:
- It's specifically designed for the task of parsing imports and exports
- 4KiB gzipped vs ~100KiB for `acorn`
- 20x faster than `acorn`

> A very small single JS file (4KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
>
> For an example of the performance, Angular 1 (720KiB) is fully parsed in 5ms, in comparison to the fastest JS parser, Acorn which takes over 100ms.
>
> Comprehensively handles the JS language grammar while remaining small and fast. - ~10ms per MB of JS cold and ~5ms per MB of JS warm, [see benchmarks](https://github.com/guybedford/es-module-lexer#benchmarks) for more info.

I gave it a test and the entire `getEsmExports` can be replaced with this:

```ts
const { init, parse } = require('es-module-lexer')

async function getEsmExports (moduleSource) {
  await init
  const srcString = moduleSource.toString()
  const [imports, exports] = parse(srcString)

  const reexports = imports
    .map(i => [srcString.slice(i.s, i.e), srcString.slice(i.ss, i.se)])
    .filter(([, full]) => full.match(/export\s*\*\s*from/))
    .map(([file]) => `* from ${file}`)

  const exportNames = exports.map(e => e.n)
  return [...exportNames, ...reexports]
}
```

A couple of test fail due to a couple of missing parser features:
https://github.com/guybedford/es-module-lexer/issues/175
https://github.com/guybedford/es-module-lexer/issues/176

BridgeAR:
I am working on that making this possible

PR #211, opened 2025-10-02
feat: Compatibility with specifier imports 

pabloerhard:
This PR adds compatibility with specifiers, for example:

`module.exports = require('#main-entry-point');`

These specifiers are defined inside the package.json file and can reference either a local file or an external module.

The proposed solution works as follows: whenever a require call begins with # (all specifiers must start with #),  locates the package.json, extracts the corresponding reference from the imports field, and sets the newUrl with the resolved path, maintains any original process after that.

### DIFF
--- a/lib/get-exports.js
+++ b/lib/get-exports.js
@@ -77,10 +77,9 @@ function resolvePackageImports (specifier, fromUrl) {
 
       currentDir = dirname(currentDir)
     }
-  } catch (error) {
-    throw new Error(`Failed to find export: ${specifier}`)
+  } catch (cause) {
+    throw Error(`Failed to find export: ${specifier}`, { cause })
   }
-
   return null
 }
 


lib/get-exports.js
```diff
@@ -27,6 +27,64 @@ function getExportsForNodeBuiltIn (name) {
...
+
+      if (existsSync(packageJsonPath)) {
+        const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
+
+        if (packageJson.imports && packageJson.imports[specifier]) {
+          const imports = packageJson.imports[specifier]
+
+          // Look for path inside packageJson
+          let resolvedPath
+          if (typeof imports == 'object') {
```
jsumners-nr, line 58: A better check is `Object.prototype.toString.call(imports) === '[object Object]'`.

jsumners-nr (approved):
Looks reasonable with linting fixed.

lib/get-exports.js
```diff
@@ -27,6 +27,64 @@ function getExportsForNodeBuiltIn (name) {
...
+            const requireSpecifier = imports.require
+            const importSpecifier = imports.import
+            // look for the possibility of require and import which is standard for CJS/ESM
+            if (requireSpecifier || importSpecifier) {
+              // trying to resolve based on order of importance
+              resolvedPath = requireSpecifier.node || requireSpecifier.default || importSpecifier.node || importSpecifier.default
```
timfish, line 64: Why does require get prioritised over import?

pabloerhard: The require field is prioritized because this resolution occurs in a CommonJS context (with require), where the require property should take priority.

lib/get-exports.js
```diff
@@ -27,6 +27,64 @@ function getExportsForNodeBuiltIn (name) {
...
+  try {
+    const fromPath = fileURLToPath(fromUrl)
+    let currentDir = dirname(fromPath)
+
+    // search for package.json file which has the real url to export
+    while (currentDir !== dirname(currentDir)) {
+      const packageJsonPath = join(currentDir, 'package.json')
+
+      if (existsSync(packageJsonPath)) {
+        const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
```
watson, line 51: Nit: It's safer to not use `existsSync` or any of its siblings, due to its inherit race conditions. Instead just call `readFileSync` and deal with the possibility that it might throw an `ENOENT` if the file doesn't exist. It's also one less I/O operation.

lib/get-exports.js
```diff
@@ -27,6 +27,64 @@ function getExportsForNodeBuiltIn (name) {
 
 const urlsBeingProcessed = new Set() // Guard against circular imports.
 
+/**
+ * This function looks for the package.json which contains the specifier trying to resolve.
+ * Once the package.json file has been found, we extract the file path from the specifier
+ * @param {*} specifier The specifier that is being search for inside the imports object
+ * @param {*} fromUrl The url from which the search starts from
```
watson, line 34:
```suggestion
 * @param {string} specifier The specifier that is being search for inside the imports object
 * @param {URL|string} fromUrl The url from which the search starts from
```

lib/get-exports.js
timfish: Why only support CJS?

AbhiPrasad (approved):

lib/get-exports.js
```diff
@@ -27,6 +27,72 @@ function getExportsForNodeBuiltIn (name) {
...
+ * Once the package.json file has been found, we extract the file path from the specifier
+ * @param {string} specifier The specifier that is being search for inside the imports object
+ * @param {URL|string} fromUrl The url from which the search starts from
+ * @returns file to url to file export
+ */
+function resolvePackageImports (specifier, fromUrl) {
+  if (!specifier.startsWith('#')) {
+    return null
+  }
+
```
BridgeAR, line 41:
```suggestion
```

This is already handled at the callsite.

lib/get-exports.js
```diff
@@ -27,6 +27,72 @@ function getExportsForNodeBuiltIn (name) {
...
+          if (resolvedPath) {
+            return pathToFileURL(require.resolve(resolvedPath, { paths: [currentDir] })).href
+          }
+        }
+        // return if we find a package.json but did not find an import
+        return null
+      }
+    }
+  } catch (error) {
```
BridgeAR, line 89:
```suggestion
  } catch {
```

Alternatively, we could also just add the `cause`

lib/get-exports.js
```diff
@@ -27,6 +27,72 @@ function getExportsForNodeBuiltIn (name) {
...
+
+      let packageJson
+      try {
+        packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
+      } catch (error) {
+        if (error.code === 'ENOENT') {
+          currentDir = dirname(currentDir)
+          continue
+        }
+        throw error
```
BridgeAR, line 58: The error itself will be discarded in the catch.

I think it would be better to not rethrow and instead break the loop and just define a state (error variable). That way this would not end up in the catch part again, which should be faster.

lib/get-exports.js
```diff
@@ -27,6 +27,72 @@ function getExportsForNodeBuiltIn (name) {
...
+              resolvedPath = requireSpecifier.node || requireSpecifier.default || importSpecifier.node || importSpecifier.default
+            } else if (imports.node || imports.default) {
+              resolvedPath = imports.node || imports.default
+            }
+          } else if (typeof imports === 'string') {
+            resolvedPath = imports
+          }
+
+          if (resolvedPath) {
+            return pathToFileURL(require.resolve(resolvedPath, { paths: [currentDir] })).href
```
BridgeAR, line 82: This is almost done identically at the callsite in the other case. Should we just define the parts and replace it with the return value? That way this part could be simplified instead of duplicating some logic.

lib/get-exports.js
```diff
@@ -27,6 +27,72 @@ function getExportsForNodeBuiltIn (name) {
...
+
+    // search for package.json file which has the real url to export
+    while (currentDir !== dirname(currentDir)) {
+      const packageJsonPath = join(currentDir, 'package.json')
+
+      let packageJson
+      try {
+        packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
+      } catch (error) {
+        if (error.code === 'ENOENT') {
```
BridgeAR, line 54: If I read the code correct, this is actually a common case. That would definitely slow down reading, since error creation is quite expensive. If we want to do that, we should likely set Error.stackTraceLimit to zero before running the readFile command. That way the error creation becomes significantly cheaper, while still being expensive.

pabloerhard: Yep, in most cases, this scenario is likely to occur, so it does makes sense to include the proposed limitations.

lib/get-exports.js
BridgeAR: That is also possible (to just return `pathToFileURL(require.resolve(re, { paths: [dirname(fileURLToPath(url))] })).href` here). I am just not sure if the method name is still appropriate in that case.

lib/get-exports.js
timfish: If this code is only hit when a specifier starts with `#`, I'm personally not that concerned about small differences in performance here. This is not going to be a common code path.

Even the race condition is unlikely to be an issue because `package.json` files should not get added or removed while this code is running.

lib/get-exports.js
pabloerhard: From what I understand, checking with toString() ensures we don't allow values such as [ ], null or Date, which are also of typeof 'object'

BridgeAR: I understand that it would not match these anymore. We just check for the property existence on these objects and they should never have the properties we are looking for. Due to that, I favor the faster check.

lib/get-exports.js
```diff
@@ -27,6 +27,63 @@ function getExportsForNodeBuiltIn (name) {
...
+            }
+          } else if (typeof imports === 'string') {
+            resolvedExport = imports
+          }
+
+          if (resolvedExport) {
+            const url = resolvedExport.startsWith('.')
+              ? pathToFileURL(join(currentDir, resolvedExport))
+              : fromUrl
+            return { url, export: resolvedExport }
```
BridgeAR, line 71: Exporting it as array would be faster :)

lib/get-exports.js
pabloerhard: I agree, I'll add that into the latest changes.

lib/get-exports.js
timfish: The best thing we could do for perf. is to migrate from `acorn` to `es-module-lexer` which is about 20x faster:
- https://github.com/nodejs/import-in-the-middle/issues/90

Unfortunately it's got a couple of outstanding bugs which are blocking migration.

timfish (approved):

lib/get-exports.js
jsumners-nr: @BridgeAR the original check was a coercive check for an `Object`. Various things, e.g `Array`, would mistakenly satisfy that check.

pabloerhard: @jsumners-nr I reinstated it to the original check after considering that the code only performs operations based on certain properties, which are checked for existence before being used. This allows us to maintain the most efficient check.

PR #219, opened 2026-01-09
test: Make "cjs" test fixture actually commonjs

isaacs:
Correct an oversight in the test fixtures added for the fix in a20f47a3013105a235f2ba48bc17319f7a57636c. The test passes fine whether the module is commonjs or esm, which is the point, but it was not actually testing the commonjs case as intended.

Make the package.json no longer define it as `"type": "module"` and convert the package modules to commonjs export format.

cc: @trentm


trentm:
Thanks!

PR #230, opened 2026-01-18
chore!: drop support for Node.js < v18

BridgeAR:
This is a breaking change, since it drops support for very old Node.js versions.


trentm:
@BridgeAR I pushed a merge from main that deals with conflicts. It might be worth a sanity check that I resolved the conflicts correctly. The conflicts were in 3 files:

```
% git merge main
Auto-merging .github/workflows/ci.yml
Auto-merging create-hook.mjs
CONFLICT (content): Merge conflict in create-hook.mjs
Auto-merging package.json
Auto-merging test/generic-loader.mjs
CONFLICT (content): Merge conflict in test/generic-loader.mjs
CONFLICT (modify/delete): test/other/get-exports-format-undefined.mjs deleted in main and modified in HEAD.  Version HEAD of test/other/get-exports-format-undefined.mjs left in tree.
Automatic merge failed; fix conflicts and then commit the result.
```

.github/workflows/ci.yml
```diff
@@ -25,13 +25,6 @@ jobs:
...
-          - 12.x
-          - 14.x
-          - 16.10.0
-          - 16.16.0
-          - 16.17.0
-          - 16.x
-          - 17.x
           - 18.5.0
           - 18.18.0
           - 18.19.0
```
aduh95: I would be dropping support for 18.x and 21.x as well, they've long been EOL

BridgeAR: That is not possible for the main users of import-in-the-middle. They have other support ranges than Node.js.

trentm: Node 18 was EOL 2025-04-30. We have a different definition of "long" ago. ;)

Just my opinion: import-in-the-middle is used, at least, by a number of observability tools that are used in enterprises where updating software versions can take years. Leaving a bit of slack when dropping support for older Node.js versions can be helpful to downstream users.  Or, from the pov of maintainers, can sometimes result in less work if supporting older Node.js versions for a while results in not ever having to do releases/maintenance on previous major versions.

Speaking for myself, downstream users of import-in-the-middle that I help maintain, and that still support Node.js 18 today are:
- `@opentelemetry/instrumentation`
- and, indirectly, `@elastic/opentelemetry-node`.

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "import-in-the-middle",
   "version": "2.0.6",
   "description": "Intercept imports in Node.js",
+  "engines": {
+    "node": ">=18"
```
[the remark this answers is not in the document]
BridgeAR: We did not do that before either and this PR does not change anything in that direction.

jsumners-nr (approved):

PR #231, opened 2026-01-19
test: increase coverage and more

BridgeAR:
I plan on separating the source fix to also reduce the PR size, since this adds a lot of test cases.


lib/get-exports.mjs
```diff
@@ -299,6 +299,21 @@ export async function getExports (url, context, parentLoad) {
   let source = parentCtx.source
   const format = parentCtx.format
 
+  // Loader hooks can return ArrayBuffer / TypedArray sources. Normalize to a
+  // string for parsing.
```
bengl: Wild. Was this a bug before?

timfish (approved):

Issue #238, opened 2026-01-24
with `internals: true` the `name` returned for the imported specifier is the implementation file, different from RITM


trentm:
Let's hook `import('got')` -- a type=module package -- using `internals: true`:

```js
import { Hook } from 'import-in-the-middle'
register('import-in-the-middle/hook.mjs', import.meta.url) 
Hook(['got'], { internals: true }, (exports, name, baseDir) => {
  console.log('HIT: name=%s', name)
})
await import('got')
```

The result:

```
HIT: name=got/dist/source/core/errors.js
...
HIT: name=got/dist/source/core/utils/is-unix-socket-url.js
HIT: name=got/dist/source/core/diagnostics-channel.js
HIT: name=got/dist/source/core/index.js
HIT: name=got/dist/source/as-promise/types.js
HIT: name=got/dist/source/as-promise/index.js
HIT: name=got/dist/source/create.js
HIT: name=got/dist/source/types.js
HIT: name=got/dist/source/index.js   <--- This one differs from RITM
```

With RITM the callback `name` value for the loaded specifier **is the specifier**, rather than the internal path. E.g. with `ioredis`:

```js
const { Hook } = require('require-in-the-middle')
Hook(['ioredis'], { internals: true }, (exports, name, baseDir) => {
  console.log('HIT: name=%s', name)
  return exports
})
require('ioredis')
```

The result:

```
HIT: name=ioredis/built/utils/lodash.js
HIT: name=ioredis/built/utils/debug.js
...
HIT: name=ioredis/built/DataHandler.js
HIT: name=ioredis/built/redis/event_handler.js
HIT: name=ioredis/built/redis/RedisOptions.js
HIT: name=ioredis/built/Redis.js
HIT: name=ioredis   <--- This one differs from IITM
```

RITM is preferring the specifier/id match over the `internals: true`-match.

### My motivation

[...]

PR #241, opened 2026-01-27
fix: ensure the callback 'name' arg is the module name when matching the module main file, even when 'internals: true' option is used

trentm:
This implements "option 1" of https://github.com/nodejs/import-in-the-middle/issues/238

Using the `got` module as an example, the issue was that the following
two hooks would have a different `name` value in the callback when the
`got` main file (".../node_modules/got/dist/source/index.js") was
loaded:

    Hook(['got'], (exports, name) => { ... })
    Hook(['got'], {internals: true}, (exports, name) => { ... })

In the first case `name === 'got'`,
in the second case `name === 'got/dist/source/index.js`.
This differed from RITM behaviour where `name === 'got'` in both cases.

This *also* fixes an issue with IITM not properly hooking an absolute
path: when the absolute path being imported happened to be in a
"node_modules" subdir.

Closes: https://github.com/nodejs/import-in-the-middle/issues/238

# Checklist

- [x] This builds on https://github.com/nodejs/import-in-the-middle/pull/237 and needs that PR to merge first.

# Details on the absolute-path import fixes

Consider this **can-iitm-match-abs-paths.mjs** which attempts to hook four absolute paths, with slightly different cases:

```js
import { register } from 'node:module'

const CWD = process.cwd()
Hook([
  `${CWD}/test/fixtures/index.js`,
  `${CWD}/test/fixtures/something.js`,
  `${CWD}/test/fixtures/node_modules/some-external-module/index.mjs`,
  `${CWD}/test/fixtures/node_modules/some-external-module/sub.mjs`
], (mod, name, baseDir) => {
  console.log(`IITM Hook: name=${name}, baseDir=${baseDir}`)
})

// 1. Import an absolute path.
await import(`${CWD}/test/fixtures/index.js`)

// 2. Import a relative path.
await import('./test/fixtures/something.js')

// 3. Absolute path, note the file happens to be in a `node_modules` dir.
await import(`${CWD}/test/fixtures/node_modules/some-external-module/index.mjs`)

// 4. Relative path, note the file happens to be in a `node_modules` dir.
await import('./test/fixtures/node_modules/some-external-module/sub.mjs')
```

With current IITM:
- Case 3 matches with a misleading (and inconsistent with other abspath matches) `baseDir` in the callback.
- Case 4 doesn't match at all.

```
% node can-iitm-match-abs-paths.mjs
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/index.js, baseDir=undefined
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/something.js, baseDir=undefined
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-module/index.mjs, baseDir=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-module
```

The reason for both issues is that the absolute files **happen to be under a `.../node_modules/...` tree**. Case 3 **accidentally** matches using the `matchArg === specifier` branch, rather than the intended match branch for absolute paths.

With the changes in this PR:

```
% node can-iitm-match-abs-paths.mjs
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/index.js, baseDir=undefined
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/something.js, baseDir=undefined
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-module/index.mjs, baseDir=undefined
IITM Hook: name=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-module/sub.mjs, baseDir=undefined
```

### Comparing to RITM

<details>
<summary>Using the equivalent test script for RITM:</summary>

**can-ritm-match-abs-paths.cjs**
```js
const { Hook } = require('require-in-the-middle')

const CWD = process.cwd()
Hook([
  `${CWD}/test/fixtures/index.js`,
  `${CWD}/test/fixtures/something.js`,
  `${CWD}/test/fixtures/node_modules/some-external-cjs-module/index.js`,
  `${CWD}/test/fixtures/node_modules/some-external-cjs-module/sub.js`
], (exports, name, baseDir) => {
  console.log(`IITM Hook: name=${name}, baseDir=${baseDir}`)
  return exports
})

// Import an absolute path.
require(`${CWD}/test/fixtures/index.js`)

// Import a relative path.
require('./test/fixtures/something.js')

// Absolute path, note the file happens to be in a `node_modules` dir.
require(`${CWD}/test/fixtures/node_modules/some-external-cjs-module/index.js`)

// Relative path, note the file happens to be in a `node_modules` dir.
require('./test/fixtures/node_modules/some-external-cjs-module/sub.js')
```

</details>

the result is:

```
% node can-ritm-match-abs-paths.cjs
IITM Hook: name=index, baseDir=/Users/trentm/tm/import-in-the-middle6/test/fixtures
IITM Hook: name=something, baseDir=/Users/trentm/tm/import-in-the-middle6/test/fixtures
IITM Hook: name=index, baseDir=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-cjs-module
IITM Hook: name=sub, baseDir=/Users/trentm/tm/import-in-the-middle6/test/fixtures/node_modules/some-external-cjs-module
```

All four match.
(Aside: the `name` and `baseDir` values are different  between IITM and RITM, but that is a topic for **a separate PR**.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,11 @@ const moduleDetailsFromPath = require('module-details-from-path')
 const { fileURLToPath } = require('url')
 const { MessageChannel } = require('worker_threads')
 
+let { isBuiltin } = require('module')
+if (!isBuiltin) {
+  isBuiltin = () => true
+}
+
 const {
   importHooks,
   specifiers,
@@ -128,7 +133,13 @@ function Hook (modules, options, hookFn) {
     let filePath, baseDir
 
     if (isNodeUrl) {
-      name = name.replace(/^node:/, '')
+      // Normalize builtin module name to *not* have 'node:' prefix, unless
+      // required, as it is for 'node:test' and some others.  `module.isBuiltin`
+      // is available in all Node.js versions that have node:-only modules.
+      const unprefixed = name.slice(5)
+      if (isBuiltin(unprefixed)) {
+        name = unprefixed
+      }
     } else if (loadUrl.startsWith('file://')) {
       const stackTraceLimit = Error.stackTraceLimit
       Error.stackTraceLimit = 0
--- a/test/hook/v18.15-node-prefixed-only.mjs
+++ b/test/hook/v18.15-node-prefixed-only.mjs
@@ -1,0 +1,12 @@
+// Test that a builtin module that must use 'node:' prefix works.
+import { strictEqual } from 'assert'
+import Hook from '../../index.js'
+
+Hook(['node:test'], (exports, name) => {
+  if (name === 'node:test') {
+    exports.skip = 'iitm was here'
+  }
+})
+
+const { skip } = await import('node:test')
+strictEqual(skip, 'iitm was here')


BridgeAR:
I did not get the chance to fully review this before it landed and I believe this change is actually not what we want, no matter that it deviates from ritm.

The reason is: if we declare `internals: true`, I really want to be on the safe side to always match the internal loaded file, no matter that the user loads it with the module name. That way it is safe that the correct code in fact always wrapped as expected. Otherwise, we would could up ending with a hook for the module name but not in case it is loaded directly by the user as internal file and I would have to declare a second hook that may only run in case the other one did not run and the other way around, if I am not mistaken (I did not yet verify this with a test case).

@trentm instead of aligning with ritm, I think we should consider changing ritm to also load the internal file when being loaded with the module name.

PR #250, opened 2026-06-02
fix: Updated proxying logic for builtins that are used as CJS

bizob2828:
At New Relic we've been in the process of migrating our instrumentation to rely on tracing channel via [orchestrion-js](https://github.com/nodejs/orchestrion-js) and the [tracing hooks](https://github.com/apm-js-collab/tracing-hooks). We noticed some incompatibilities when registering both IITM and `@apm-js-collab/tracing-hooks`.  I was able to distill it down to edge cases of built in packages.  

 When IITM wraps a Node built-in (e.g. node:events, diagnostics_channel),
  CJS modules that call require() on it via Node 22+'s require(esm) path
  would receive either the raw IITM proxy or undefined, causing:
    - TypeError: Class extends value [object Module] is not a constructor
    - TypeError: Cannot destructure property 'tracingChannel' of undefined

  Four fixes:
  - Explicitly add 'module.exports' to the builtin exports set on Node >= 23,
    where it is a synthetic named ESM export for CJS modules
  - Keep track of cjs modules imported and do not wrap so when requiring module it returns the appropriate export
  - Add a $default fallback in the generated shim for the module.exports
    binding, so that when namespace['module.exports'] is undefined (Node
    doesn't expose it for all builtins), the actual CJS module.exports value
    is used instead
  - Use Object.getOwnPropertyNames (instead of Object.keys) when building
    the exports set for builtins, so non-enumerable properties are included

This was all observed when creating an esm app and importing koa. Since our agent rewrites koa files with orchestrion-js, it was producing the crashes above.

EDIT:

I still had an issue on node 22 and could not figure out how to solve. I used claude and it came up with a fix and also some good tests that mimic the behavior between iitm, tracing hooks and a package like koa. This was its summary of the fix:

```
 ---
  Root Cause

  Two issues compound to cause ERR_VM_MODULE_LINK_FAILURE:

  1. Node.js bug (Node 22+): When a CJS module is loaded through the ESM translator and a loader hook provides its source
  (instead of leaving it null for Node to read natively), require() calls inside that CJS module fail for packages using the
   "module-sync" exports condition (like generator-function). The module-sync condition resolves to an ESM file
  (require.mjs), which imports CJS files — and ModuleJob.runSync can't link those dependencies when hook-provided source
  changes the CJS loading path.

  2. Missing propagation in cjsInIitmChain: IITM tracked top-level CJS modules it wrapped, but didn't propagate the
  "in-chain" status to their transitive require() targets. So koa/lib/application.js was in the set, but its children
  (is-generator-function, call-bound, etc.) were not.

  Fix (in create-hook.mjs)

  1. Resolve hook — Propagate cjsInIitmChain membership to child modules so the entire synchronous CJS require tree is
  tracked:
  if (cjsInIitmChain.has(parentURL)) {
    cjsInIitmChain.add(result.url) // NEW
    return result
  }
  2. Load hook — Strip hook-provided source for CJS modules in the chain, forcing Node to use its native CJS loader which
  handles module-sync correctly:
  if (cjsInIitmChain.has(url)) {
    const result = await parentLoad(url, context)
    if (result.format === 'commonjs' && result.source != null) {
      return { format: result.format, source: undefined }
    }
    return result
  }

  Test Files Added

  - test/hook/v22-cjs-transitive-require-chain.mjs — Tests cjsInIitmChain propagation
  - test/hook/v22-cjs-require-module-sync-with-source-hook.mjs — Tests the full scenario with a source-providing hook (run
  with --import test/fixtures/register-iitm-with-source-hook.mjs)
  - Supporting fixtures in test/fixtures/
 ```

### DIFF
--- a/create-hook.mjs
+++ b/create-hook.mjs
@@ -423,7 +423,11 @@ export function createHook (meta) {
     // When a CJS module is loaded by an IITM shim, its require() calls for
     // builtins may be routed through the ESM resolver on Node 24+. Skip IITM
     // wrapping in that case so require() returns the native module value.
+    // We also propagate the membership to the resolved child so that its own
+    // transitive require() calls are likewise skipped (the entire synchronous
+    // CJS require chain must remain unwrapped to avoid ERR_VM_MODULE_LINK_FAILURE).
     if (cjsInIitmChain.has(parentURL)) {
+      cjsInIitmChain.add(result.url)
       return result
     }
 
@@ -588,6 +592,24 @@ register(${JSON.stringify(realUrl)}, _, set, get, ${JSON.stringify(originalSpeci
 
       // Fall back to the parent loader with the original (non-iitm) URL.
       return parentLoad(deleteIitm(url), context)
+    }
+
+    // On Node 22+, when a CJS module is loaded through the ESM translator and
+    // another loader hook provides its source (instead of leaving source null
+    // for Node to read natively), require() calls inside that CJS module for
+    // packages using the "module-sync" exports condition fail with
+    // ERR_VM_MODULE_LINK_FAILURE. Work around this Node bug by stripping
+    // hook-provided source for CJS modules in the synchronous require chain,
+    // forcing Node to use its native CJS loader which handles this correctly.
+    if (cjsInIitmChain.has(url)) {
+      const result = await parentLoad(url, context)
+      if (result.format === 'commonjs' && result.source != null) {
+        return {
+          format: result.format,
+          source: undefined
+        }
+      }
+      return result
     }
 
     return parentLoad(url, context)
--- a/create-hook.mjs
+++ b/create-hook.mjs
@@ -265,12 +265,19 @@ async function processModule ({ srcUrl, context, parentGetSource, parentResolve,
       const objectKey = JSON.stringify(n)
       const reExportedName = n === 'default' ? n : objectKey
 
+      // For the module.exports synthetic export (Node 23+), fall back to $default
+      // when namespace['module.exports'] is not exposed by the native ESM namespace
+      // (builtins don't expose it). This ensures the IITM hook proxy returns the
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: test/hook/v23-builtin-module-exports-shim.mjs [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+import Hook from '../../index.js'
+
+let capturedExports = null
+const hook = new Hook(['events'], (exports, name) => {
+  if (name === 'events') {
+    capturedExports = exports
+  }
+})
+
+const eventsNs = await import('node:events')
```
isaacs, line 22: It would probably be good to verify that this is the thing we expect, if only just to make sure we don't break it later?

bizob2828: yea I moved this into draft because I thought I had fixes in place but they broke some of the integration tests.  I'll add an assertion here as well

create-hook.mjs
```diff
@@ -308,6 +315,14 @@ export function createHook (meta) {
   const iitmURL = new URL('lib/register.js', meta.url).toString()
   let includeModules, excludeModules
 
+  // Track CJS module URLs that IITM has wrapped. On Node 24+, CJS modules loaded
+  // via loadCJSModule (in an ESM import chain) have their require() calls for
+  // builtins routed through the ESM resolver. Without this guard, IITM would
+  // intercept those require() calls and return an ESM namespace object instead
+  // of the native CJS module value (e.g. EventEmitter constructor), breaking
+  // patterns like `class App extends require('events') {}`.
```
jsumners-nr, line 323: Just... amazing code that.

jsumners-nr (approved):
I have not reviewed all of the tests. I did read through the code, well, mainly the comments. Seems reasonable to me.

timfish (approved):

PR #255, opened 2026-06-18
feat: add shouldInclude predicate for consumer-owned module matching

BridgeAR:
## Summary

`include`/`exclude` only accept bare specifiers, file URLs, and regular expressions. Consumers that need different matching — a matcher built from their own config, a prefix/glob scheme, or a decision based on more than the specifier — have to encode everything into those shapes or run a parallel filter outside iitm.

This adds an optional `shouldInclude(url, specifier)` predicate. When provided, it replaces the include/exclude lists: iitm calls it for each resolved module and the consumer owns the decision, keeping its matching in one place. The predicate receives the resolved URL and the specifier; iitm does not resolve a file path for it (the list matcher still does, for absolute-path entries added via `Hook` over the message port).

`module.register` structured-clones its `data`, so a function can't be passed through it — the predicate is for synchronous registration (`module.registerHooks` / `register-hooks.mjs`) or a predicate built on the loader thread. `include`/`exclude` remain fully supported when no predicate is given.

As a drive-by, the `result.format` guard is hoisted to the top of `finishResolve` so it applies to both the predicate and list paths (a predicate can legitimately match a non-JS format such as `node_modules/<pkg>/package.json`).

## Test plan

- [ ] New `test/register/v22.15-sync-register-hooks-should-include.mjs`: the predicate gates wrapping and is called with `(url, specifier)` only.
- [ ] Existing include/exclude and `v18.19-include-message-port-absolute-path` tests pass unchanged (list path untouched).

### DIFF
--- a/create-hook.mjs
+++ b/create-hook.mjs
@@ -352,7 +352,8 @@ function addIitm (url) {
 export function createHook (meta) {
   let cachedResolve
   const iitmURL = new URL('lib/register.js', meta.url).toString()
-  let includeModules, excludeModules, shouldInclude
+  let includeModules, excludeModules
+  let shouldInclude = defaultShouldInclude
 
   // Track CJS module URLs that IITM has wrapped. On Node 24+, CJS modules loaded
   // via loadCJSModule (in an ESM import chain) have their require() calls for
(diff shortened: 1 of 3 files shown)


timfish (dismissed):

create-hook.mjs
```diff
@@ -430,38 +442,42 @@ export function createHook (meta) {
...
-    function match (each) {
-      if (each instanceof RegExp) {
-        return each.test(result.url)
+    if (shouldInclude !== undefined) {
+      // Consumer-owned matcher: it decides from the URL and specifier alone, so
+      // the resolved file path is neither computed nor passed here.
+      if (!shouldInclude(result.url, specifier)) {
+        return result
       }
+    } else {
```
bengl: Maybe this whole bit should just be a default shouldInclude, rather than branching here?

BridgeAR: Done

timfish (approved):

rochdev (approved):

jsumners-nr (approved):

PR #258, opened 2026-06-25
fix: instrument type-stripped TypeScript modules

### LINKED ISSUE #186: Support for --experimental-strip-types

BridgeAR:
## Summary

When `--experimental-strip-types` is active, Node reports `.ts`/`.mts`/`.cts`
files as `module-typescript` / `commonjs-typescript`. Those formats were absent
from `HANDLED_FORMATS`, so the resolve hook returned them uninstrumented — hooks
never fired.

Node hands the `load` hook the original un-stripped source, so the export parser
strips before parsing. The TS formats are added to `HANDLED_FORMATS` only when
`stripTypeScriptTypes` is available (Node >= 22.13), because a static named import
of it throws on older runtimes; runtimes that pre-date the API continue to skip TS
modules instead of crashing.

## Test plan

- [ ] `test/hook/v24-typescript-hook.mjs` covers both `module-typescript` (`.mts`)
  and `commonjs-typescript` (`.cts`); it asserts the hook fires and captures the
  correct export names after type stripping.
- [ ] The test skips automatically below Node v24, so the suite is safe across the
  full version matrix.

Fixes: https://github.com/nodejs/import-in-the-middle/issues/186

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,34 @@ registration (`register-hooks.mjs`, shown above) and for predicates constructed 
 the loader thread; it is not accepted through the `data` option of the asynchronous
 `module.register('import-in-the-middle/hook.mjs', ...)`.
 
+## TypeScript modules
+
+On Node.js versions that strip TypeScript types natively (those exposing
+[`module.stripTypeScriptTypes()`](https://nodejs.org/api/module.html#modulestriptypescripttypescode-options),
+>= 22.13.0 / >= 23.9.0 / >= 24.0.0), `import-in-the-middle` intercepts `.ts`,
+`.mts` and `.cts` modules just like their JavaScript counterparts. The types are
+stripped before the module's exports are read, so type-only exports
+(`export type`, `export interface`) are not present on the intercepted
+namespace; value exports are.
+
+```ts
+// math.mts
+export type Op = '+' | '-'
+export function add (a: number, b: number): number {
+  return a + b
+}
+```
+
+```js
+Hook(['./math.mts'], (exported) => {
+  // `exported.add` is interceptable; the `Op` type is not part of the namespace
+})
+```
+
+On Node.js versions where type stripping is not enabled by default, run with
+`--experimental-strip-types`. Older versions that predate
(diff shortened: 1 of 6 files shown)


jsumners-nr (approved):
🤷‍♂️

timfish (approved):

PR #265, opened 2026-06-30
perf: read each wrapped module from disk once

BridgeAR:
Wrapping a module reads its source twice and re-lexes shared leaves once per
barrel. Two changes remove both, with no observable behavior change:

1. The export pass reads a module's source to lex its exports; the wrapper it
   emits then does `import * as namespace from <realUrl>`, so Node reads the
   same file again to compile that import. Stash the source the export pass
   fetched and serve it to the namespace import.
2. A leaf reached from several `export *` barrels is lexed once per barrel even
   though its export set is intrinsic to the file. Memoize the lexed ESM export
   names by URL so the first lex serves the rest.

On a 390-module barrel-heavy graph under the synchronous loader: export lexes
588 → 396 (-33%), total module read calls 1970 → 794 (-60%).

The source cache only holds entries whose upstream load supplied a format; a
load that returns source without one relies on Node to resolve the format on
the real load, and serving such an entry would throw ERR_UNKNOWN_MODULE_FORMAT.
The export memo only caches the pure ESM result — the CommonJS path mutates
`context.format` and resolves re-exports per call, and built-ins already
memoize via BUILT_INS.

### DIFF
--- a/create-hook.mjs
+++ b/create-hook.mjs
@@ -7,7 +7,8 @@ import { inspect } from 'util'
 import { builtinModules } from 'module'
 import {
   getExports,
-  hasModuleExportsCJSDefault
+  hasModuleExportsCJSDefault,
+  takeCachedSource
 } from './lib/get-exports.mjs'
 import { RESOLVE, driveSync, driveAsync } from './lib/io.mjs'
 import { supportsSyncHooks } from './supports-sync-hooks.mjs'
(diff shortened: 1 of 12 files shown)


create-hook.mjs
```diff
@@ -218,43 +218,27 @@ function buildSetter (n, srcUrl) {
...
+  // the actual CJS value (e.g. EventEmitter) rather than undefined. `1` selects
+  // that fallback; every other export passes nothing.
+  const fallback = n === 'module.exports' ? ', 1' : ''
+
+  // Each export keeps the irreducible pair the module system requires — a live
+  // `let` binding and a static `export { ... }` — and hands the binder a
+  // reader/writer for it. All the read / defer / override / set / get wiring
+  // lives once in `__bind` (see buildWrapperSource), so the generated source is
+  // a couple of lines per export instead of ~25; the wrapper Node must compile
+  // shrinks accordingly. The module.exports re-export is still dropped for
```
bengl: Shrinks compared to what? This comments makes more sense as a commit message.

bengl (approved):

jsumners-nr (approved):

PR #266, opened 2026-07-01
fix: keep get-esm-exports synchronous so require() works

BridgeAR:
## Summary

`es-module-lexer`'s parser selection awaited its result at the top level, which made `lib/get-esm-exports.mjs` an async ES module. That poisons the whole chain that pulls it in — `create-hook.mjs` and the register-hooks entry a CommonJS preloader `require()`s to install the synchronous loader — so `require()` throws `ERR_REQUIRE_ASYNC_MODULE` and the synchronous `module.registerHooks` loader can't be registered from a `require()` context. A consumer (e.g. dd-trace) that `require()`s its loader entry silently falls back to the off-thread loader.

The top-level await was never necessary. `es-module-lexer`'s main entry resolves to a CommonJS build via its `require` export condition, and the `es-module-lexer/js` asm.js subpath is CommonJS too, so both parser builds load synchronously through `createRequire`. Selecting the parser synchronously drops the top-level await and keeps the module `require()`-able, without changing which build is used on either path.

## Why

The `--disallow-code-generation-from-strings` branch previously came in via `await import('es-module-lexer/js')` on the assumption that the asm.js build had no CommonJS variant. It does — the subpath's `default` condition points at a plain-CJS `dist/lexer.asm.js` — so `require()` loads it just as well, synchronously.

## Test plan

- [x] `test/get-esm-exports/require-sync-get-esm-exports.mjs` — new regression test: `require()` of `get-esm-exports.mjs` must not throw `ERR_REQUIRE_ASYNC_MODULE`, and the parser still lexes exports and `export *` re-exports. Fails on `main` with `ERR_REQUIRE_ASYNC_MODULE`, passes with this change.
- [x] Same regression test under `NODE_OPTIONS=--disallow-code-generation-from-strings` to cover the asm.js branch.
- [x] `test/get-esm-exports/*` and `test/register/*` (incl. the `v22.15-sync-register-hooks*` suites) pass.

### DIFF
--- a/lib/get-esm-exports.mjs
+++ b/lib/get-esm-exports.mjs
@@ -1,6 +1,9 @@
 'use strict'
 
+import { createRequire } from 'module'
 import { initSync, parse as parseWasm } from 'es-module-lexer'
+
+const require = createRequire(import.meta.url)
 
 // The default es-module-lexer build is WebAssembly backed and decodes string
 // literals (import specifiers, quoted export names) with an internal `eval`.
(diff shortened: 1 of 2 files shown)


BridgeAR:
This is somewhat a hack now. I am working on exposing cjs directly from es-module-lexer as well. This is just the in-between workaround.

timfish (approved):

bengl (approved):

PR #281, opened 2026-08-10
feat: allow disabling CJS source stripping

pabloerhard:
## Summary

Adds a `disableCjsSourceStripping` hook option to opt out of IITM’s CJS source stripping workaround.

By default, IITM strips hook provided source from CJS modules in its synchronous require chain. This avoids `ERR_VM_MODULE_LINK_FAILURE` when a CJS module requires a package that uses the `"module-sync"` exports condition.

However, the workaround causes every subsequent synchronous hook to receive `undefined` as the module source. While that resolves the affected `module-sync` cases, it prevents downstream synchronous hooks from processing CJS modules that do not have this issue.

Consumers that need a downstream loader hook to receive and process CJS source can now disable this behavior:

```js
register({ disableCjsSourceStripping: true })
```

The option works with both synchronous (`register-hooks.mjs`) and asynchronous (`module.register(...)`) registration.

## Tests

- Added async coverage for the default source-stripping behavior.
- Added async and sync coverage for the opt-out behavior.
- Ran `npm test`.

## Notes
The ERR_VM_MODULE_LINK_FAILURE error that the original workaround addresses is not consistently present across all Node.js versions. Node 24.18.0+ does not seem to suffer from it, however Node 22 and 26 latest does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,33 @@ registration (`register-hooks.mjs`, shown above) and for predicates constructed 
 the loader thread; it is not accepted through the `data` option of the asynchronous
 `module.register('import-in-the-middle/hook.mjs', ...)`.
 
+### Opting out of CJS source stripping
+
+On Node 22+, when a CJS module is loaded through the ESM translator and another
+loader hook provides its source (instead of leaving it `null` for Node to read
+natively), `require()` calls inside that CJS module for packages using the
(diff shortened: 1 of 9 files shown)


BridgeAR (commented):
Mind adding this option to the type definition as well? :)

LGTM otherwise

BridgeAR (approved):

jsumners-nr (approved):

Issue #171, opened 2024-12-13
Duplicate exports from the same source file are incorrectly excluded 


timfish:
If a module has multiple exports with the same name, these do not get exported. However, if a duplicate named exports comes from the same source file, it should be retained. 

A simple reproduction:
`a.mjs`
```ts
export const val = 1;
```

`b.mjs`
```ts
export * from './a.mjs';
```

`c.mjs`
```ts
export * from './a.mjs';
export * from './b.mjs';
```
When run in Node, `c.mjs` has a single export named `val`.
When running through `import-in-the-middle` there is no export.

This issues currently affects for the following libraries:
- `svelte`
- `@langchain/core`
- `drizzle-orm`

Using these libraries when they have been wrapped by `import-in-the-middle` results in errors like:
- `ReferenceError: X is not defined`
- `The requested module 'X' does not provide an export named 'Y'`

wouterds:
@dependabot merge

PR #282, opened 2026-08-18
fix: resolve 'ambiguous' chained `export *` via explicit fallback

### LINKED ISSUE #269: Breaks newer typebox (export * behaviour?)
I believe something is mishandling the extensive use of `export *` in https://github.com/sinclairzx81/typebox/blob/0c31e10b1d080ed95c2a58e52aaf8c517b7aaf18/src/type/types/index.ts , but don't understand enough about either project to provide a better guess.

```bash
echo '{"type":"module"}' > package.json
npm i import-in-the-middle typebox
```

`repro.js`
```js
import * as Type from 'typebox/type';
import { Value } from 'typebox/value';
import { Type as T } from 'typebox';

console.log('IsTemplateLiteral:', typeof Type.IsTemplateLiteral);
console.log('IsString:', typeof Type.IsString);
console.log('IsObject:', typeof Type.IsObject);

try {
  const unionSchema = T.Object({ level: T.Union([T.Literal('a'), T.Literal('b')]) });
  console.log('Convert union OK:', JSON.stringify(Value.Convert(unionSchema, { level: 'a' })));
} catch (e) {
  console.log('CONVERT union ERROR:', (e).message);
}
```

```
$ node repro.js
IsTemplateLiteral: function
IsString: function
IsObject: function
Convert union OK: {"level":"a"}
```

```
$ node --loader=import-in-the-middle/hook.mjs repro.js
...
IsTemplateLiteral: undefined
...
CONVERT union ERROR: Type.IsTemplateLiteral is not a function
```

```
├── import-in-the-middle@3.3.1
└── typebox@1.3.6
```

FauxFaux:
Fixes #269. Partially reworks #171 / #263.

Full disclosure: This change was significantly AI assisted, hence the extensive comment churn and unit tests. I do believe it to be reasonable, but am far from a domain expert. It is additive, and does not change any existing tests, so I believe it is worth raising. This body is fully human.

---

When there are two levels of `export *`, and a cyclic import, the previous code would handle this as if there was a truly ambiguous double `export *`, and remove the export (as specified). This differs from the nodejs (and presumably specification) behaviour, where it is allowable to fallback to the original defining(?) namespace.

This is broken into two commits: one attempts to re-work the fix in #171 by making this fallback behaviour explicit, then the second extends the fallback to apply (instead of erasure) in the typebox case.

This second commit, while ugly in the diff, is mostly just moving code around to handle an extra `if` condition, not adding or removing any weird code. It is now valid to do this because of the new fallback behaviour. Focus on this `addSetter` change.

### DIFF
--- a/create-hook.mjs
+++ b/create-hook.mjs
@@ -198,21 +198,25 @@ function emitWarning (err) {
  * of how the loader is driven, so both the synchronous and asynchronous paths
  * share it.
  *
- * The value is read from `namespaceVar`, the wrapper's namespace binding for the
- * module that *defines* the export. For a module's own exports that is the
- * wrapped module itself; for a name re-exported through `export *` it is the
- * leaf that declares it. Reading from the defining module rather than the
- * aggregating one keeps the value resolvable when the same binding reaches the
- * aggregator through more than one re-export chain — Node sees those chains as
- * distinct wrapper modules and leaves the name ambiguous (hence `undefined`) on
- * the aggregate namespace, while the defining module always holds it (#171).
+ * The value is read from `namespaceVar`, the wrapped module's own namespace. A
+ * name the same binding reaches through more than one `export *` chain
+ * additionally passes `fallbackVar`, the namespace of the module that *defines*
+ * it: the binder reads the aggregate namespace first (so a hook applied along
+ * the re-export chain is still observed) and falls back to the defining module
+ * when the aggregate does not hold the value — Node sees the chains as distinct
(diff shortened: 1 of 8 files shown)


FauxFaux:
A further review has found issues with this approach, notably that the fallback module is not hooked.

Which doesn't matter at all for my usecase (typescript with opentelemetry, where there's no instrumenter inside typescript), but is a real bug we probably shouldn't plaster over.

